PR #7, opened 2025-12-15
Avoiding clone for some operations, introducing CoW, improving code

cc90202:
Please consider to add this operations to have tests successfully passed:

  docker exec mosaico-testing-mosaico-pg-testing-1 psql -U postgres -d mosaico \
    -c "DROP SCHEMA public CASCADE; CREATE SCHEMA public;"

and then:
  rm -rf /tmp/__mosaico_auto_testing__/*

Otherwise the dirty data will not allow to pass tests

### DIFF
--- a/mosaicod/src/main.rs
+++ b/mosaicod/src/main.rs
@@ -117,7 +117,7 @@ fn run(startup_time: &Instant) -> Result<(), Box<dyn std::error::Error>> {
                     args.host,
                     args.port,
                     &store_display_name,
-                    &get_version(),
+                    get_version(),
                     startup_time,
                 );
             })?;
--- a/mosaicod/src/repo/sql_models/topic_record.rs
+++ b/mosaicod/src/repo/sql_models/topic_record.rs
@@ -67,8 +67,7 @@ impl TopicRecord {
     pub fn serialization_format(&self) -> Option<rw::Format> {
         self.serialization_format
             .as_ref()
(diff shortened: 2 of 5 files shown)


cabba:
Hi Cristiano, thanks for the great contribution!

Before merging I have a couple of questions: 
- Did `clippy` reported some errors/warnings for you? In my local env was all fine
- Are you seeing any errors running tests (script `scripts/test_suite.sh`)? Because the script has a cleanup phase in which the docker compose is torn down using `-v` option (so volumes are destroyed) and the temporary directory in `/tmp` is removed, [link](https://github.com/mosaico-labs/mosaico/blob/b82b799d58b888ce9b80fcb3ad3d19b08ad9a466/scripts/test_suite.sh#L82C10-L93)

cc90202:
Hi, 
1) my clippy is pedantic, I've set a lot of warning that usually are not set by default. During my updates some warning has raises, but likely from my code  and then I've removed them.
2) Yes. 3 tests didn't pass (don't remember which of them). I've investigated and the reason should be that there are dirty data in docker and DB . So I've make a clean and then they passed all.

cabba:
Thanks! The PR has been merged, but I need to investigate why the cleanup phase of the test suite script didn’t run. Could you let me know which machine you’re using and your Docker version?

cc90202:
Pc windows 11, wsl2,  Docker version 28.5.0, build 887030f

PR #13, opened 2025-12-16
Fix/async mixed with sync

cc90202:
This PR fixes async blocking issues and introduces several performance optimizations for the mosaicod server.

  Changes

  Performance

  - Replace system allocator with mimalloc - Optimize memory allocation performance in concurrent workloads (Arrow Flight server, parallel chunk processing) with  throughput improvement
  - Parallelize chunk query processing - Use FuturesUnordered with semaphore-based concurrency limiting. Pre-fetch all required topics upfront
  - Offload CPU-intensive parquet operations to blocking threads - Wrap parquet write/finalize operations with spawn_blocking to prevent async executor blocking during ZSTD compression
  - Reduce unnecessary clones and allocations - Use &'static str instead of String, Cow<'static, str> for TextStats, &str parameters where possible

  Configuration

  - Add configurable parameters for db pool and chunk concurrency - New environment variables (env.develop): MOSAICO_MAX_DB_CONNECTIONS (default: 10) and MOSAICO_MAX_CONCURRENT_CHUNK_QUERIES (default: 4)

  Build & Infrastructure (toml)

  - Add release profile optimizations - LTO, strip symbols, single codegen unit, panic=abort for optimal runtime performance
  - Remove unused benchmark infrastructure - Clean up criterion dependency to reduce compile times

  Code Quality

  - Improve code quality based on rust guidelines - Use #[expect] with reason, replace unwrap() with expect(), handle network interface errors gracefully

### DIFF
--- a/mosaicod/Cargo.lock
+++ b/mosaicod/Cargo.lock
@@ -2450,6 +2450,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f9fbbcab51052fe104eb5e5d351cf728d30a5be1fe14d9be8a3b097481fb97de"
 
 [[package]]
+name = "libmimalloc-sys"
+version = "0.1.44"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "667f4fec20f29dfc6bc7357c582d91796c169ad7e2fce709468aefeb2c099870"
+dependencies = [
+ "cc",
+ "libc",
+]
+
+[[package]]
 name = "libredox"
 version = "0.1.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


cc90202:
Few words to explain the commit f56ce92.
In the previous code you had:
 1.  Find all chunks that could have data
  2. For each chunk, check if data match the filter
  3. Return the relevant chunk only
  
 Chunk 1: [----10ms----]
                       Chunk 2: [----10ms----]
                                                            Chunk 3: [----10ms----]
                                                                          ...

With FuturesInorderd:
  Default (in env.develop) MOSAICO_MAX_CONCURRENT_CHUNK_QUERIES=4.
  Time:         |----10ms----         |      ----10ms----                          |----10ms----| .....
  Chunk 1:   [============]
  Chunk 2:   [============]
  Chunk 3:   [============]
  Chunk 4:   [============]
                                                   Chunk 5:   [============]
                                                   Chunk 6:   [============]

mosaicod/src/rw/chunked_writer.rs
```diff
@@ -110,7 +110,16 @@ where
             None => ChunkWriter::try_new(batch.schema(), self.format)?,
         };
 
-        writer.write(batch)?;
+        // Clone batch for spawn_blocking (requires 'static)
+        let batch = batch.clone();
+
+        // Offload CPU-intensive parquet encoding/compression to blocking thread pool
+        writer = tokio::task::spawn_blocking(move || {
+            writer.write(&batch)?;
```
cabba: I think that the best approach here is to make the writer.write async using [`AsyncArrowWriter`](https://arrow.apache.org/rust/parquet/arrow/async_writer/struct.AsyncArrowWriter.html).
It should be an easy fix but I didn't have the time to look into it.

PR #37, opened 2025-12-20
fix(sdk-py): correct __exit__ return type for mypy

AndreaBozzo:
mypy catches the bool as an error, propagation behaviour unchanged, low priority for merge
- Fix mypy `exit-return` error: change `__exit__` return type from `bool` to `None`

### DIFF
--- a/mosaico-sdk-py/src/mosaicolabs/comm/mosaico_client.py
+++ b/mosaico-sdk-py/src/mosaicolabs/comm/mosaico_client.py
@@ -170,12 +170,11 @@ class MosaicoClient:
         exc_type: Optional[Type[BaseException]],
         exc_val: Optional[BaseException],
         exc_tb: Optional[Any],
-    ) -> bool:
+    ) -> None:
         """
         Context manager exit point. Ensures resources are closed.
 
-        Returns:
-            bool: False, to propagate any exceptions raised within the `with` block.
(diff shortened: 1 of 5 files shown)


AndreaBozzo:
## Update: Extended fix to all handlers

Added same fix to 4 more files with identical pattern:
- `sequence_handler.py`
- `topic_handler.py`
- `topic_writer.py`
- `sequence_writer.py`

## Rationale

Per mypy best practices, `__exit__` return type should be:
- `-> None` when exceptions are **never suppressed** (our case)
- `-> bool` only if exceptions **might be suppressed** (returning `True`)

Using `-> bool` when always returning `False` is misleading—it implies the context manager could swallow exceptions, affecting mypy's reachability analysis.

**Sources:**
- [mypy error codes - exit-return](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-exit-does-not-return-a-value-exit-return)
- [Adam Johnson - How to type a context manager](https://adamj.eu/tech/2021/07/04/python-type-hints-how-to-type-a-context-manager/)
- [typeshed #5073 - Fix __exit__ annotation](https://github.com/python/typeshed/issues/5073)

fdicorato:
I agree with the changes, given the current version of the code.
(In the first versions of the code the __exit__ methods returned True in some paths. Then these returns were removed)

PR #39, opened 2025-12-22
Refactor/code review

cc90202:
Summary

  This PR improves code quality and maintainability through targeted refactoring based on comprehensive code review analysis. The changes reduce cyclomatic complexity, enforce proper Rust idioms.

  Changes

  - Extract do_action handlers - Split monolithic do_action.rs (complexity: 85) into dedicated handler modules (actions/sequence.rs, actions/topic.rs, actions/layer.rs, actions/query.rs), reducing per-handler complexity to ~15
  - Apply Strategy Pattern for Format - Implemented FormatStrategy trait to handle format-specific Parquet configuration (Default, Ragged, Image), fixing Open/Close principle violation and eliminating ignored parameters
  - Use #[expect] with reasons - Replaced #[allow] with #[expect] including justification comments per M-LINT-OVERRIDE-EXPECT guideline
  - Replace unsafe with _force suffix - Renamed functions like topic_delete → topic_delete_force since "dangerous operations" (data loss) ≠ undefined behavior, following M-UNSAFE-IMPLIES-UB
  - Propagate runtime build error - Replaced .unwrap() with ? in server startup to properly handle tokio runtime creation failures

### DIFF
--- a/mosaicod/src/server/endpoints/actions/mod.rs
+++ b/mosaicod/src/server/endpoints/actions/mod.rs
@@ -1,0 +1,31 @@
+//! Action handlers for Flight DoAction requests.
+//!
+//! This module contains free functions for handling Flight actions,
+//! organized by resource type (sequence, topic, layer, query).
+
+pub mod layer;
+pub mod query;
+pub mod sequence;
+pub mod topic;
+
+use crate::{query as ts_query, repo, store};
+
+/// Shared context for all action handlers.
+///
+/// Contains references to the store, repository, and timeseries engine
+/// that handlers need to perform their operations.
+pub struct ActionContext {
+    pub store: store::StoreRef,
+    pub repo: repo::Repository,
+    pub ts_gw: ts_query::TimeseriesGatewayRef,
+}
+
+impl ActionContext {
+    pub fn new(
+        store: store::StoreRef,
+        repo: repo::Repository,
+        ts_gw: ts_query::TimeseriesGatewayRef,
+    ) -> Self {
+        Self { store, repo, ts_gw }
+    }
+}
--- a/mosaicod/src/server/endpoints/do_action.rs
+++ b/mosaicod/src/server/endpoints/do_action.rs
@@ -1,310 +1,79 @@
-use log::{info, trace, warn};
+//! Flight DoAction endpoint implementation.
+//!
+//! This module implements the main dispatcher for Flight DoAction requests,
+//! delegating to specialized handler functions for each action category.
 
 use crate::{
-    marshal::{self, ActionRequest, ActionResponse},
-    query,
-    repo::{self, FacadeError, FacadeLayer, FacadeQuery, FacadeSequence, FacadeTopic},
+    marshal::{ActionRequest, ActionResponse},
+    query, repo,
     server::errors::ServerError,
-    store, types,
-    types::{MetadataBlob, Resource},
+    store,
 };
 
+use super::actions::{ActionContext, layer, query as query_action, sequence, topic};
+
+/// Dispatches a Flight action request to the appropriate handler.
+///
+/// This function serves as the main entry point for all Flight DoAction requests,
+/// routing each action type to its specialized handler function.
 pub async fn do_action(
     store: store::StoreRef,
     repo: repo::Repository,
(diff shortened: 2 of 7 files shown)


mosaicod/src/rw/format.rs
```diff
@@ -1,7 +1,173 @@
...
+/// - Closed for modification: Existing code doesn't need to change when adding formats
+pub trait FormatStrategy: Send + Sync {
+    /// Returns the file extension for this format (without leading dot).
+    fn file_extension(&self) -> &'static str;
+
+    /// Returns the Parquet writer properties configured for this format.
+    fn writer_properties(&self) -> WriterProperties;
+
+    /// Returns DataFusion ListingOptions configured for reading files in this format.
+    fn listing_options(&self) -> ListingOptions;
```
cabba, line 50: In future updates, `Format` will be used also to define non-Parquet data formats (like point clouds or images), so we will probably need to use something other than DataFusion with those formats.

I like the use of the strategy pattern here, but we need to think about how to handle `listing_options` in cases in which it is not required. Maybe we can define an additional `ListingOptions` trait?

mosaicod/src/server/endpoints/actions/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+
+/// Shared context for all action handlers.
+///
+/// Contains references to the store, repository, and timeseries engine
+/// that handlers need to perform their operations.
+pub struct ActionContext {
+    pub store: store::StoreRef,
+    pub repo: repo::Repository,
+    pub ts_engine: ts_query::TimeseriesGwRef,
+}
```
cabba, line 21: Very good idea

mosaicod/src/server/endpoints/do_action.rs
```diff
@@ -1,312 +1,99 @@
...
-            let loc = handle.locator.clone();
-            handle.delete().await?;
-            warn!("resource {} deleted", loc);
-
-            ActionResponse::Empty
+    match action {
+        // Sequence actions
+        ActionRequest::SequenceCreate(data) => {
+            let user_metadata = data.user_metadata()?;
+            SequenceActionHandler::create(&ctx, data.name, user_metadata.as_str()).await
```
cabba: I’m concerned that introducing `SequenceActionHandler` (and all the other handlers) might add unnecessary abstraction, especially since we already have `Facade` implementations (like `FacadeSequence` and `FacadeTopic`) handling these elements. I worry that adding another layer might make the logic a bit harder to follow.

As an alternative to keep cyclomatic complexity low, what do you think about using free functions? We could house them in a module like `actions/sequence.rs` with straightforward functions for `create`, `delete`, etc. This might keep the implementation leaner while still achieving the decoupling we're after.

mosaicod/src/rw/format.rs
cc90202: The solution I can see is: 
1) defining a trait with the minimal interface that all formats must satisfied:
 ```
pub trait FormatStrategy: Send + Sync {
      fn file_extension(&self) -> &'static str;
      fn name(&self) -> &'static str;
  }
```

Then a Parquet specific:
```
pub trait ParquetFormat: FormatStrategy {
      fn writer_properties(&self) -> WriterProperties;
      fn listing_options(&self) -> ListingOptions;
  }
```

Future: other format families:
```
pub trait PointCloudFormat: FormatStrategy {
      fn point_cloud_reader_config(&self) -> PointCloudReaderConfig;
  }

  pub trait RawImageFormat: FormatStrategy {
      fn image_codec(&self) -> ImageCodec;
  }
```

Ragarding Format enum:

```
 match format {
      Format::Default | Format::Ragged | Format::Image => {
          // use ParquetFormat trait
      }
      Format::PointCloud(_) => {
          // different read path entirely
      }
  }
```

mosaicod/src/server/endpoints/do_action.rs
cc90202: Free functions is a good idea.

mosaicod/src/rw/format.rs
cabba: I think is a good future-proof architecture

cc90202:
I must rebase and align to new implementations due to conflicts.

PR #40, opened 2025-12-22
feat(server): implement list_flights endpoint for catalog discovery

### LINKED ISSUE #27: Enable catalog discovery via empty path in `list_flights`
### Summary
Implement `list_flights` endpoint to support root-level catalog discovery. When queried with an empty path (`""`) or a root path (`"/"`), the server should return a `ListFlightStream` object containing a list of all available sequences in the platform.

### Motivation
The goal of this task is to implement a new behavior for `list_flights` so that when a client requests info using an empty path (`""`) or a root path (`"/"`), the server responds with streamed list of all available sequences in the platform.

While it would be possible to return all `SequenceTopicGroups` for the entire platform in response to an empty query in `query` action, I think it will be better to return only the list of sequences with a dedicated endpoint to maintain optimal performance and avoid heavy metadata processing.

### Proposed Solution
- Introduce a new logic `src/server/endpoints/list_flights.rs`.
- Add an associated function to `FacadeSequence` to fetch a list of all sequence resource locators currently stored in the system.
- Map each discovered sequence to a `FlightEndpoint`. Each endpoint will contain a `Ticket` derived from the sequence name, allowing the client to make subsequent requests for specific sequence details
- To maintain high performance, the root response will only list sequence identifiers. Fetching full topic-level metadata for every sequence in a single call is excluded to avoid database bottlenecks (since `#topic >> #sequence`).

AndreaBozzo:
## Summary
- Implement Arrow Flight `list_flights` endpoint for root-level catalog discovery
- Stream all sequences when querying with empty or root path (`""` or `"/"`)
- Return minimal `FlightInfo` (identifier only) to avoid metadata overhead

Closes #27

## Changes
- `mosaicod/src/server/endpoints/list_flights.rs` - New endpoint implementation
- `mosaicod/src/server/flight.rs` - Wire up endpoint in FlightService
- `mosaicod/src/repo/mod.rs` - Export `sequence_find_all` query

## Notes
Edits enabled due to this being critical for next release, feel free to add inline tests ( i can also do this ) or adjust implementation as needed.

### DIFF
--- a/mosaicod/src/server/endpoints/list_flights.rs
+++ b/mosaicod/src/server/endpoints/list_flights.rs
@@ -7,7 +7,7 @@ use futures::stream::BoxStream;
 use log::{info, trace};
 use tonic::Status;
 
-use crate::{repo, server::errors::ServerError};
+use crate::{repo, server::errors::ServerError, types::Resource};
 
 /// Lists all available flights (sequences) in the repository.
 ///


mosaicod/src/server/endpoints/list_flights.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    if !is_root_query {
+        return Err(ServerError::UnsupportedDescriptor);
+    }
+
+    info!("listing all sequences");
+
+    // Fetch all sequences from repository
+    let mut cx = repo.connection();
+    let sequences = repo::sequence_find_all(&mut cx).await?;
```
cabba: I will wrap this in a associated method of the `FacadeSequence`, something like 

```rust
pub fn async FacadeSequence::all() -> Vec<types::SequenceResourceLocator>
```

mosaicod/src/server/endpoints/list_flights.rs
```diff
@@ -0,0 +1,64 @@
...
+
+use crate::{repo, server::errors::ServerError, store};
+
+/// Lists all available flights (sequences) in the repository.
+///
+/// When clients query with an empty or root path ("" or "/"), this function
+/// returns a streamed list of all sequences. Each sequence is represented
+/// as a minimal `FlightInfo` containing only the sequence identifier.
+pub async fn list_flights(
+    _store: store::StoreRef,
```
cabba: If it is not used we can remove it, in order to be more explicit interface about the resources used.

AndreaBozzo: +1 for removing it

AndreaBozzo:
Had to push the facade change aswell, you have full access to the branch if needed

PR #51, opened 2026-01-02
fix(stats): use Option for TextStats min/max to handle empty strings correctly

AndreaBozzo:
- Fix critical bug in `TextStats` where empty string `""` used as placeholder caused incorrect min/max calculation when `""` was a legitimate value
- Refactor `sanitize_name` to handle multiple leading slashes
- Add unit tests for edge cases

## Problem
When a dataset contained an empty string as a valid value (e.g., `["", "a"]`), the min would incorrectly be calculated as `"a"` instead of `""` because the empty placeholder couldn't be distinguished from actual empty values.

## Solution
Changed `TextStats.min` and `TextStats.max` from `Cow<'static, str>` to `Option<String>`:

## Test plan
- [x] Added 4 unit tests covering: empty string as min, normal values, null-only data, merge with empty string
- [x] All 25 unit tests pass
- [x] Clippy passes with no warnings

### DIFF
--- a/mosaicod/src/types/chunk.rs
+++ b/mosaicod/src/types/chunk.rs
@@ -1,4 +1,3 @@
-use std::borrow::Cow;
 use std::collections::HashMap;
 
 const NUMERIC_MIN_PLACEHOLDER: f64 = f64::MAX;
@@ -99,8 +98,8 @@ impl NumericStats {
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct TextStats {
-    pub min: Option<Cow<'static, str>>,
-    pub max: Option<Cow<'static, str>>,
+    pub min: Option<String>,
+    pub max: Option<String>,
 
     pub has_null: bool,
 }
@@ -127,13 +126,13 @@ impl TextStats {
         if let Some(val) = val {
             let val = *val;
             match &self.min {
-                Some(current_min) if **current_min <= *val => {}
-                _ => self.min = Some(Cow::Owned(val.to_owned())),
+                Some(current_min) if current_min.as_str() <= val => {}
+                _ => self.min = Some(val.to_owned()),
             }
 
             match &self.max {
-                Some(current_max) if **current_max >= *val => {}
-                _ => self.max = Some(Cow::Owned(val.to_owned())),
+                Some(current_max) if current_max.as_str() >= val => {}
+                _ => self.max = Some(val.to_owned()),
             }
         } else {
             self.has_null = true;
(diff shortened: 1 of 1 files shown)


mosaicod/src/types/chunk.rs
```diff
@@ -105,8 +99,8 @@ impl NumericStats {
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct TextStats {
-    pub min: Cow<'static, str>,
-    pub max: Cow<'static, str>,
+    pub min: Option<Cow<'static, str>>,
+    pub max: Option<Cow<'static, str>>,
```
cabba: Using both `Option` and `Cow` seems a bit redundant.

We can use `Option<String>` with `None` as uninitialized string (as your proposal), or `Cow<'static, str>` using `Cow::Borrowed` variant to express uninitialization.

Probabily I prefer `Option<String>` since it is a bit simpler and clearer, what do you think?

AndreaBozzo: Thanks for the review! I agree that Option<String> is simpler and cleaner and we can refactor to use that.

Regarding the Cow::Borrowed("") alternative for expressing uninitialization: unfortunately that approach would reintroduce the original bug. The problem is that using an empty string as a sentinel value conflates two distinct states:

- Uninitialized: Cow::Borrowed("")
- Legitimate empty string value: Cow::Borrowed("") or Cow::Owned("".to_string())

Both would be represented identically, so we'd lose the ability to correctly handle datasets like ["", "a"] where "" is a valid minimum value. This is a [known antipattern](https://jwilm.io/blog/from-str-to-cow/) ( also looking for better sources/examples about this), using sentinel values instead of Option to represent absence. The idiomatic Rust approach is to use Option<T> for optional values, keeping Cow orthogonal for the borrowed/owned optimization concern. With Option<String>:

- None = uninitialized (no values seen yet)
- Some("".to_string()) = empty string is the actual min/max

Since our use case always ends up owning the string (via to_owned() in eval() and merge()), the Cow optimization doesn't provide real benefits here anyway.

leaving this here for feedbacks before eventually refactor, also open if mantainers want to handle that.

cabba: Thanks for the interesting read.

Wat I meant is that since we have that borrowed state happens only in initialization ([link](https://github.com/mosaico-labs/mosaico/blob/408b4cf6ea1f5591df6743f08c5ecd1bdc8e5604/mosaicod/src/types/chunk.rs#L123-L124)) we can check if the string is borrowed at the first iteration and assign an owned string just after, something like:
```rust
if self.min.is_borrowed() {
    self.min = Cow::Owned(...)
}
```
with the `Borrowed` variant expressing the uninitialized condition. At this point there is no ambiguity since:

- **Uninitialized**: Cow::Borrowed("")
- **Legitimate empty string**: Cow::Owned("".to_string())

That said, I agree that is clearer to just use `Option` since there is no real benefit in performance and using `Cow` keeps things a little bit more obscure and requires prior knowledge to understand the behaviour.

**Note**: I've just realized that `is_borrowed` is an experimental API in `Cow` so that justifies even more the use of `Option`.

If you have time to fix that would be great, otherwise we can handle this.

AndreaBozzo: Pushed fixes in commit 6c152fc

PR #52, opened 2026-01-04
perf(resources): optimize SequenceTopicGroups::merge O(n*m) → O(n+m)

AndreaBozzo:
- Replace linear search with HashMap lookup for O(1) sequence matching
- Eliminate `.clone()` by consuming the second group with `remove()`
- Add test case for merge behavior

This is an optional optimization, open to feedback on the approach.

## References
- [The Rust Performance Book - Hashing](https://nnethercote.github.io/perf-book/hashing.html)

### DIFF
--- a/mosaicod/examples/merge_benchmark.rs
+++ b/mosaicod/examples/merge_benchmark.rs
@@ -7,11 +7,11 @@ use std::time::{Duration, Instant};
 
 // Standalone replica of project types
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 #[allow(dead_code)]
 struct TopicResourceLocator(String);
 
-#[derive(Clone, PartialOrd, Ord, PartialEq, Eq)]
+#[derive(Clone, PartialOrd, Ord, PartialEq, Eq, Debug)]
 struct SequenceResourceLocator(String);
 
 impl SequenceResourceLocator {
@@ -20,7 +20,7 @@ impl SequenceResourceLocator {
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 struct SequenceTopicGroup {
     sequence: SequenceResourceLocator,
     topics: Vec<TopicResourceLocator>,
@@ -75,12 +75,12 @@ impl SequenceTopicGroups {
 
         group
             .0
-            .sort_by(|a, b| a.sequence.name().cmp(&b.sequence.name()));
+            .sort_by(|a, b| a.sequence.name().cmp(b.sequence.name()));
 
         for mut self_grp in self.0 {
             let found = group
                 .0
-                .binary_search_by(|grp_aux| self_grp.sequence.name().cmp(&grp_aux.sequence.name()));
+                .binary_search_by(|grp_aux| self_grp.sequence.name().cmp(grp_aux.sequence.name()));
 
             if let Ok(found) = found {
                 self_grp.topics.extend(group.0[found].topics.clone());


AndreaBozzo:
I had this laying on my fork for a few days, sending it now before major changes by mantainers, have a nice sunday everyone

cabba:
I think we need to measure performance here. For small sets, simple linear searches can outperform hash map lookups due to cache locality and avoiding hash computations and collisions.

The best approach would be to check performance on a dataset similar to the real one, with topics in the form `sequence_name/a/b/c`.

Edit: in our case, the number of topics will be roughly between ten and a thousand.

AndreaBozzo:
> I think we need to measure performance here. For small sets, simple linear searches can outperform hash map lookups due to cache locality and avoiding hash computations and collisions.
> 
> The best approach would be to check performance on a dataset similar to the real one, with topics in the form `sequence_name/a/b/c`.
> 
> Edit: in our case, the number of topics will be roughly between ten and a thousand.

I ran a standalone benchmark to compare the two implementations. The benchmark measures pure merge time (clone overhead subtracted) across different dataset sizes.

**Test machine:** Intel Core i3-4330 @ 3.50GHz (2C/4T), 8GB RAM

mergetopicbench


- Linear wins for small datasets (< 100 sequences) due to cache locality
- HashMap wins at 100+ sequences where O(n*m) complexity starts to dominate
- The crossover point is around 50-100 sequences on my machine

Given that the expected range is 10-1000 sequences, both implementations have their sweet spot. If most real-world usage falls in the lower end (< 100), linear might be preferable. If usage is more varied or tends toward the higher end, HashMap is the better choice. The benchmark is available at examples/merge_benchmark.rs if you want to test on other machines.


I'm also ok to converting this to a discussion

AndreaBozzo:
Also confirming workflows to run on PRs now

cabba:
Did some testing and rebalanced the logic between sequences and topics; topics are more likely to scale (10 to 1k), while sequence scaling will be query-dependent.

I implemented a sort + binary search optimization which seems like the right middle ground for this case.

Screenshot From 2026-01-05 16-10-08


One catch: I noticed that the HashMap version behaves slightly differently than the linear one: if a sequence group contains the same sequence twice, which is possible in the original structure, the merge will overwrite one of them.

AndreaBozzo:
I ran the updated benchmark locally and got some anomalous results:

```
    seqs   topics        hashmap         linear        ratio
------------------------------------------------------------
       5       10        4.24 us           0 ns      1.00x =
      10       20           0 ns           0 ns      1.00x =
      20       50           0 ns           0 ns      1.00x =
      50      100      257.83 us           0 ns      1.00x =
      80      500        1.74 ms        1.15 ms      0.66x L
     100     1000        4.68 ms        3.90 ms      0.83x L
```

The `0 ns` values indicate that the clone overhead subtraction is causing issues, when the clone time measurement has high variance, `saturating_sub` truncates negative results to zero.

Also, the current "linear" implementation is no longer O(n*m), it uses sort + binary_search which is O(n log n + m log m). This is a different algorithm than the original linear search we were comparing against.

To get reliable measurements, we should either:
1. Remove clone subtraction and measure total time
2. Pre-allocate separate copies for each iteration before timing

Personal note:
This is proving to be deeper than initially expected, i have to personally advice against this change as is right now.

**Test machine:** Intel Core i3-4330 @ 3.50GHz (2C/4T), 8GB RAM
mergetopicbinary

Edit: Added screen

cabba:
> Also, the current "linear" implementation is no longer O(n*m), it uses sort + binary_search which is O(n log n + m log m). This is a different algorithm than the original linear search we were comparing against.

Yes, exactly, precisely $(n+m) \log (m)$. It is a classic cache-friendly second step optimization over linear searches. Since we are dealing with small sets keeping data in the hot path is tipically more effective that computational complexity.

> The 0 ns values indicate that the clone overhead subtraction is causing issues, when the clone time measurement has high variance, saturating_sub truncates negative results to zero.

Ops, I've missed the part in the code where the time was subtracted. Sorry.

> To get reliable measurements, we should either:
>
> - Remove clone subtraction and measure total time
> - Pre-allocate separate copies for each iteration before timing

I belive removing the clone subtraction is safe here since we are measuring the same clone in both methods and the cost is also averaged over 2000 iterations.

cabba:
> Personal note:
> This is proving to be deeper than initially expected, i have to personally advice against this change as is right now.

Since we’ve already done some work on this, it would be a shame to lose it. We can keep working on it as a low-priority task for fun :smile:.

AndreaBozzo:
Worst case scenario is a good lesson learned!

Can keep It for visibility and more benches later on, maybe someone else gives It a shot

PR #65, opened 2026-01-05
Implement Scalar Ontology Models: Temperature, Pressure and Range

### LINKED ISSUE #60: Implement Scalar Ontology Models: Temperature, Pressure and Range
As part of the effort to expand the Mosaico data ontology, this issue focuses on implementing the first tier of scalar sensor models in the Python SDK. These models - Temperature, Pressure, and Range - are considered "good first contributions" as they follow the established patterns of existing *base* types.

**Key Requirements**
* SI Units: Strictly adhere to SI units to ensure consistency for downstream Physical AI models:
   * Kelvin for Temperature;
   * Pascal for Pressure;
   * Meters for Range;
* Schema Design: Define PyArrow-compatible structs for each model.
* Metadata Integration:
   * Inherit from HeaderMixin to ensure proper time stamping and frame identification;
   * Inherit from CovarianceMixin to ensure proper covariance injection.
* ROS Alignment: Ensure ontology types are compatible with common ROS message definitions (see the list) to facilitate data injection through the ros_bridge.
   * https://docs.ros2.org/latest/api/sensor_msgs/msg/FluidPressure.html
   * https://docs.ros2.org/latest/api/sensor_msgs/msg/Range.html
   * https://docs.ros2.org/latest/api/sensor_msgs/msg/Temperature.html

NikBellini:
- Implement Temperature, Pressure and Range Ontology Models
- Add tests for the three Ontology Models

Closes #60

### DIFF
--- a/mosaico-sdk-py/src/mosaicolabs/models/sensors/range.py
+++ b/mosaico-sdk-py/src/mosaicolabs/models/sensors/range.py
@@ -9,11 +9,10 @@ from typing_extensions import Self
 from pydantic import model_validator
 
 from ..header_mixin import HeaderMixin
-from ..covariance_mixin import CovarianceMixin
 from ..serializable import Serializable
 
 
-class Range(Serializable, HeaderMixin, CovarianceMixin):
+class Range(Serializable, HeaderMixin):
     """
     Range measurement data.
     """
@@ -25,9 +24,7 @@ class Range(Serializable, HeaderMixin, CovarianceMixin):
                 "radiation_type",
                 pa.uint8(),
                 nullable=False,
-                metadata={
-                    "description": "Which type of radiation the sensor used."
-                },
+                metadata={"description": "Which type of radiation the sensor used."},
             ),
             pa.field(
                 "field_of_view",
@@ -59,9 +56,7 @@ class Range(Serializable, HeaderMixin, CovarianceMixin):
                 "range",
                 pa.float32(),
                 nullable=False,
-                metadata={
-                    "description": "Range value in meters."
-                },
+                metadata={"description": "Range value in meters."},
             ),
         ]
     )
--- a/mosaico-sdk-py/src/mosaicolabs/models/sensors/pressure.py
+++ b/mosaico-sdk-py/src/mosaicolabs/models/sensors/pressure.py
@@ -7,11 +7,10 @@ Defines the data structure for pressure sensors.
 import pyarrow as pa
 
 from ..header_mixin import HeaderMixin
-from ..covariance_mixin import CovarianceMixin
 from ..serializable import Serializable
 
 
-class Pressure(Serializable, HeaderMixin, CovarianceMixin):
+class Pressure(Serializable, HeaderMixin):
     """
     Pressure measurement data.
     """
@@ -23,7 +22,9 @@ class Pressure(Serializable, HeaderMixin, CovarianceMixin):
                 "fluid_pressure",
                 pa.float64(),
                 nullable=False,
-                metadata={"description": "The absolute pressure reading from the sensor in Pascals."},
+                metadata={
+                    "description": "The absolute pressure reading from the sensor in Pascals."
+                },
             ),
             pa.field(
                 "variance",
--- a/mosaico-sdk-py/src/mosaicolabs/models/sensors/temperature.py
+++ b/mosaico-sdk-py/src/mosaicolabs/models/sensors/temperature.py
@@ -7,11 +7,10 @@ Defines the data structure for temperature sensors.
 import pyarrow as pa
 
 from ..header_mixin import HeaderMixin
-from ..covariance_mixin import CovarianceMixin
 from ..serializable import Serializable
 
 
-class Temperature(Serializable, HeaderMixin, CovarianceMixin):
+class Temperature(Serializable, HeaderMixin):
     """
     Temperature measurement data.
     """
--- a/mosaico-sdk-py/src/testing/unit/query/test_query_sensors.py
+++ b/mosaico-sdk-py/src/testing/unit/query/test_query_sensors.py
@@ -469,8 +469,6 @@ class TestQueryTemperatureAPI:
         Temperature.Q.header.stamp.sec
         Temperature.Q.header.stamp.nanosec
         Temperature.Q.header.frame_id
-        # Inherited from CovarianceMixin
-        Temperature.Q.covariance_type
         # Inherited from Message
         Temperature.Q.timestamp_ns
         # --- Catalog Context: Non-existing field ---
@@ -486,7 +484,6 @@ class TestQueryTemperatureAPI:
         # Local fields
         assert issubclass(type(Temperature.Q.temperature), _QueryableNumeric)
         assert issubclass(type(Temperature.Q.variance), _QueryableNumeric)
-        assert issubclass(type(Temperature.Q.covariance_type), _QueryableNumeric)
         assert issubclass(type(Temperature.Q.header.seq), _QueryableNumeric)
         assert issubclass(type(Temperature.Q.header.stamp.sec), _QueryableNumeric)
         assert issubclass(type(Temperature.Q.header.stamp.nanosec), _QueryableNumeric)
@@ -560,8 +557,6 @@ class TestQueryPressureAPI:
         Pressure.Q.header.stamp.sec
         Pressure.Q.header.stamp.nanosec
         Pressure.Q.header.frame_id
-        # Inherited from CovarianceMixin
-        Pressure.Q.covariance_type
         # Inherited from Message
         Pressure.Q.timestamp_ns
         # --- Catalog Context: Non-existing field ---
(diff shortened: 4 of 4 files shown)


mosaico-sdk-py/src/mosaicolabs/models/sensors/range.py
```diff
@@ -0,0 +1,83 @@
+"""
```
AndreaBozzo, line 1: Docstring refers another module

AndreaBozzo:
@NikBellini Very nice so far, i just added a minor suggestion, feel free to tag  if needed!

NikBellini:
@AndreaBozzo thank you! Just fixed

AndreaBozzo:
> @AndreaBozzo thank you! Just fixed

You can set the PR to open so we can check the workflows aswell

fdicorato:
Hi @NikBellini, thank you for your contribution (and speed). 

Here are my reviews:
* remove `variance` from class field and pyarrow schema in `Pressure` and `Temperature`: they both inherit from `CovarianceMixin` which is the standard mosaico way of injecting uncertainty fields
* naming: maybe we can rename the `Pressure.fluid_pressure` and `Temperature.temperature` to `value` or `data`? @cabba @AndreaBozzo @NikBellini  what do you think?
* `Range`:
   * missing `type` field, which for now we can keep as `int`, defining the type of range measurement (acoustic or light-based)
   * The main doubt is about keeping the fields `field_of_view`, `min_range` and `max_range` in the class definition, since they can be assumed as sensor specific (intrinsic parameters), that may not change during a single mission. ROS defines these parameters inside the message for design, but in our case they can be assumed as topic-specific metadata; see discussion #61 also. Please, tell me what you think about

NikBellini:
Hi @fdicorato, I will immediately remove the `variance` and will add the `type` field to the `Range` Model.

About the naming, I agree with you. The current names are very redundant and the model name already describes what is the value contained.

About the fields `field_of_view`, `min_range` and `max_range`, I think it's a good idea, avoiding saving repeated data.

AndreaBozzo:
>     * naming: maybe we can rename the `Pressure.fluid_pressure` and `Temperature.temperature` to `value` or `data`? @cabba @AndreaBozzo @NikBellini  what do you think?
>

Semantics vs uniformity

>     * `Range`:
>       
>       * missing `type` field, which for now we can keep as `int`, defining the type of range measurement (acoustic or light-based)
>       * The main doubt is about keeping the fields `field_of_view`, `min_range` and `max_range` in the class definition, since they can be assumed as sensor specific (intrinsic parameters), that may not change during a single mission. ROS defines these parameters inside the message for design, but in our case they can be assumed as topic-specific metadata; see discussion [Improved mapping of ROS message constants and hardware intrinsic parameters #61](https://github.com/mosaico-labs/mosaico/discussions/61) also. Please, tell me what you think about

This should stay in #61 as it would heavily change the range module

AndreaBozzo:
>     * remove `variance` from class field and pyarrow schema in `Pressure` and `Temperature`: they both inherit from `CovarianceMixin` which is the standard mosaico way of injecting uncertainty fields
> 

Will the  `ros_bridge` map variance → covariance[0]? Genuine question

NikBellini:
For now I added the `radiation_type` to the `Range` model for uniformity as mentioned by @AndreaBozzo.  Once and if it is decided to change names, I will change them all

NikBellini:
Convert this pull request to draft for now

fdicorato:
> > ```
> > * naming: maybe we can rename the `Pressure.fluid_pressure` and `Temperature.temperature` to `value` or `data`? @cabba @AndreaBozzo @NikBellini  what do you think?
> > ```
> 
> Semantics vs uniformity
> 
> > ```
> > * `Range`:
> >   
> >   * missing `type` field, which for now we can keep as `int`, defining the type of range measurement (acoustic or light-based)
> >   * The main doubt is about keeping the fields `field_of_view`, `min_range` and `max_range` in the class definition, since they can be assumed as sensor specific (intrinsic parameters), that may not change during a single mission. ROS defines these parameters inside the message for design, but in our case they can be assumed as topic-specific metadata; see discussion [Improved mapping of ROS message constants and hardware intrinsic parameters #61](https://github.com/mosaico-labs/mosaico/discussions/61) also. Please, tell me what you think about
> > ```
> 
> This should stay in #61 as it would heavily change the range module

I totally agree. Let's keep the fields and we will refactor once #61 is closed

fdicorato:
> > ```
> > * remove `variance` from class field and pyarrow schema in `Pressure` and `Temperature`: they both inherit from `CovarianceMixin` which is the standard mosaico way of injecting uncertainty fields
> > ```
> 
> Will the `ros_bridge` map variance → covariance[0]? Genuine question

The mixin is designed to represent the n*n (n dimension of the uncertain components) flattened (co)variance matrix. I undrestand that variance → covariance[0] could be seen as a degenerate case, but I still would prefer uniform design with respect to the other types. Maybe defining a VarianceMixin (scalar variance) on the side of CovarianceMixin can be more acceptable?

cabba:
> naming: maybe we can rename the Pressure.fluid_pressure and Temperature.temperature to value or data?

I agree, `value` is pretty clear to me.

AndreaBozzo:
Regarding the `variance` vs `CovarianceMixin` discussion:

@fdicorato's idea of a **VarianceMixin** for scalar sensors is elegant and would solve the semantic mismatch cleanly:

- `CovarianceMixin` → n×n covariance matrices (IMU, Odometry, etc.)
- `VarianceMixin` → scalar variance (Temperature, Pressure, etc.)

This would also clarify how `ros_bridge` should handle the mapping.

**Suggestion:** Should we open a separate issue/discussion to design `VarianceMixin`? This PR could proceed with `CovarianceMixin` only (as requested), and scalar sensors could migrate to `VarianceMixin` once it's implemented.

This keeps the scope of this PR focused while tracking the architectural improvement separately.

fdicorato:
> Regarding the `variance` vs `CovarianceMixin` discussion:
> 
> @fdicorato's idea of a **VarianceMixin** for scalar sensors is elegant and would solve the semantic mismatch cleanly:
> 
> * `CovarianceMixin` → n×n covariance matrices (IMU, Odometry, etc.)
> * `VarianceMixin` → scalar variance (Temperature, Pressure, etc.)
> 
> This would also clarify how `ros_bridge` should handle the mapping.
> 
> **Suggestion:** Should we open a separate issue/discussion to design `VarianceMixin`? This PR could proceed with `CovarianceMixin` only (as requested), and scalar sensors could migrate to `VarianceMixin` once it's implemented.
> 
> This keeps the scope of this PR focused while tracking the architectural improvement separately.

Sure! I'll create the issue with the proposed schema for approval
Thank you!

NikBellini:
I added the validator for checking that `min_range` is less or equal to `max_range` and another validator that checks that `range` stays between `min_range` and `max_range`, as the documentation says that the `range` < `min_range` or `range` > `max_range` should be discarded:

https://docs.ros2.org/latest/api/sensor_msgs/msg/Range.html

Let me know if the second validator is ok or should be removed and handled by the user.

For the moment I left the `variance` field. In case, I can work on the `VarianceMixin` issue as it's strictly connected to this PR. Once completed I will remove the `variance` field

cabba:
> Please, rewrite in your own words the comments in the `"description":` metadata values (pyarrow schemas) copied from ROS library with no attribution, as by doing so you may be breaking licensing terms. 
> 
> 
> 
> *Please, do not use attribution, rewrite them in other words.*

We need to write this as a warning somewhere in a discussion @fdicorato

fdicorato:
> > Please, rewrite in your own words the comments in the `"description":` metadata values (pyarrow schemas) copied from ROS library with no attribution, as by doing so you may be breaking licensing terms.
> > _Please, do not use attribution, rewrite them in other words._
> 
> We need to write this as a warning somewhere in a discussion @fdicorato

We have it here:
https://github.com/mosaico-labs/mosaico/blob/main/IP.md
 > "*NO Copy-Pasting*"

However it is ok to write a post highlighting this part of the documentation

NikBellini:
I fixed the descriptions. Sorry, I didn't know that even copying only the description could cause problems (I have read the "NO Copy-Pasting"). Hope it's ok now.

About the `CovarianceMixin`, because there is the `variance` field, should I remove it for now until the `VarianceMixin` or should I remove the `variance` field?

fdicorato:
> I fixed the descriptions. Sorry, I didn't know that even copying only the description could cause problems (I have read the "NO Copy-Pasting"). Hope it's ok now.
> 
> About the `CovarianceMixin`, because there is the `variance` field, should I remove it for now until the `VarianceMixin` or should I remove the `variance` field?



> I fixed the descriptions. Sorry, I didn't know that even copying only the description could cause problems (I have read the "NO Copy-Pasting"). Hope it's ok now.
> 
> About the `CovarianceMixin`, because there is the `variance` field, should I remove it for now until the `VarianceMixin` or should I remove the `variance` field?



> I fixed the descriptions. Sorry, I didn't know that even copying only the description could cause problems (I have read the "NO Copy-Pasting"). Hope it's ok now.
> 
> About the `CovarianceMixin`, because there is the `variance` field, should I remove it for now until the `VarianceMixin` or should I remove the `variance` field?

No problem, thank you! 
Yes, let's leave the `variance` field for keeping the PR consistent. 
The `VarianceMixin` will be available by tomorrow morning at most. Then i will open the issue for changing the classes in this PR.

PR #68, opened 2026-01-07
chore(ci): split workflow into parallel jobs

### LINKED ISSUE #57: Improve action workflow
The current GitHub Actions workflow bundles multiple checks into a single action, which makes failures harder to interpret and slows down iteration. 
Splitting the workflow into dedicated phases would improve readability, debuggability, and overall CI feedback.

We should break the workflow into separate jobs for:
- [ ] Backend linter
- [ ] Backend unit tests
- [ ] Python SDK unit tests
- [ ] Integration tests

This would allow parallelization and more targeted improvements to each phase.

AndreaBozzo:
This should align the workflow with modern devops practices
- Splits the monolithic `lint-and-test` job into 4 independent jobs for better parallelization and debugging
- Renames workflow from `tests.yml` to `ci.yml`

**New job structure:**
| Job | Dependencies | Content |
|-----|--------------|---------|
| `backend-linter` | none | Rust linting with clippy |
| `backend-unit-tests` | none | Rust unit tests |
| `python-sdk-unit-tests` | none | Python SDK unit tests |
| `integration-tests` | all above | End-to-end tests |

## Test plan
- [ ] Verify all 4 jobs run correctly on this PR (it won't, this PR renames tests.yml)
- [ ] Confirm first 3 jobs run in parallel
- [ ] Confirm integration-tests waits for others to complete

## References
- [Swatinem/rust-cache - shared-key](https://github.com/Swatinem/rust-cache)
- [actions/setup-python Poetry caching](https://github.com/actions/setup-python/issues/846)
- [GitHub Actions caching best practices](https://medium.com/@amareswer/github-actions-caching-and-performance-optimization-38c76ac29151)
- [Optimizing GitHub Actions Workflows 2025](https://marcusfelling.com/blog/2025/optimizing-github-actions-workflows-for-speed)

To be monitored

Closes #57

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,7 @@ permissions:
 
 jobs:
   backend-linter:
+    name: "Backend (linter)"
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Code
@@ -51,26 +52,8 @@ jobs:
         run: cargo lint
 
   backend-unit-tests:
+    name: "Backend (unit tests)"
     runs-on: ubuntu-latest
-    services:
-      postgres:
-        image: postgres:16
-        env:
-          POSTGRES_USER: postgres
-          POSTGRES_PASSWORD: password
-          POSTGRES_DB: mosaico
-        ports:
-          - 6543:5432
-        options: >-
-          --health-cmd pg_isready
-          --health-interval 10s
-          --health-timeout 5s
-          --health-retries 5
-    env:
-      DATABASE_URL: postgresql://postgres:password@localhost:6543/mosaico
-      MOSAICO_REPOSITORY_DB_URL: postgresql://postgres:password@localhost:6543/mosaico
-      SQLX_OFFLINE: true
-      RUST_BACKTRACE: 1
     steps:
       - name: Checkout Code
         uses: actions/checkout@v6
@@ -97,6 +80,7 @@ jobs:
         run: ./scripts/tests.sh --mosaicod
 
   python-sdk-unit-tests:
+    name: "Python SDK (unit tests)"
     runs-on: ubuntu-latest
     env:
       PYTHON_KEYRING_BACKEND: keyring.backends.null.Keyring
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,200 @@
...
+        run: |
+          mkdir -p /tmp/__mosaico_auto_testing__
+          cd mosaicod
+          ./target/debug/mosaicod run --port 6276 --local-store /tmp/__mosaico_auto_testing__ > /tmp/mosaicod_e2e_testing.out 2>&1 &
+          MOSAICOD_PID=$!
+          echo "Started mosaicod with PID: $MOSAICOD_PID"
+          sleep 5
+          cd ../mosaico-sdk-py
+          poetry run pytest ./src/testing -k integration
+          kill $MOSAICOD_PID 2>/dev/null || true
```
cabba: Maybe we can put this into a script? In order to test the same stuff when we run locally.

AndreaBozzo: This is the trickiest part, i am also thinking about a separated script like run_integration_tests.sh, the current one excutes all lint + tests + int tests.

AndreaBozzo: Testcontainers is another alternative for this, but adds dependencies

AndreaBozzo:
> Maybe we can split this single workflow into 2 different workflows:
> 
>     * One for the backend that runs cargo test and integration tests
> 
>     * One for the python sdk that runs python unit tests and integration tests
> 
> 
> What do you think?

to add on this, int tests require **both** , this points directly on how we currently handle integration tests repo-wide

AndreaBozzo:
I ended up adding the dedicated script, we can improve CI/CD with future iterations as the project expands

cabba:
> to add on this, int tests require both , this points directly on how we currently handle integration tests repo-wide

We’re fine for now, but as we scale, the wait times will get annoying. We shouldn't be stuck waiting on backend tests if we’re just making internal SDK changes, and integration testing will fail for any API change.

AndreaBozzo:
> > to add on this, int tests require both , this points directly on how we currently handle integration tests repo-wide
> 
> We’re fine for now, but as we scale, the wait times will get annoying. We shouldn't be stuck waiting on backend tests if we’re just making internal SDK changes, and integration testing will fail for any API change.

This can be further discussed btw, i'm very up to keep it simple

cabba:
I’ve been thinking about this, and I believe the best approach would be to update the existing `test_suite.sh` script (and maybe rename it to `tests.sh`). We can modify it to accept optional arguments like this:

```bash
# Spins up docker testing env and runs mosaicod unit tests
./test.sh --mosaicod   
# Runs poetry commands and python sdk unit tests (no docker env)
./test.sh --sdk-python 
# Spins up docker testing env, runs poetry, and runs integration tests
./test.sh --integration 
# Runs all of the above (default behavior)
./test.sh --all
```
This way, we can remove the postgres setup in `ci.yml` and also selectively run individual steps during development and on CI workflows, what do you think?

Let me know if you want to handle it or if I should take care of it.

cabba:
Thanks!

PR #102, opened 2026-01-20
feat(mosaicod): implement transactional outbox pattern for S3 uploads

cc90202:
Summary                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                    
  This PR implements the Transactional Outbox pattern to eliminate S3 orphan files that could occur when the database insert fails after a successful S3 upload.                                                                                                                    
                                                                                                                                                                                                                                                                                    
  Problem                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                    
  The previous write path in chunked_writer.rs performed a non-atomic dual-write:                                                                                                                                                                                                   
  1. Write Parquet file to S3                                                                                                                                                                                                                                                       
  2. Insert metadata in PostgreSQL                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                    
  If step 2 failed after step 1 succeeded, the S3 file became an orphan with no corresponding database record.                                                                                                                                                                      
                                                                                                                                                                                                                                                                                    
  Solution                                                                                                                                                                                                                                                                          
       
 Reverse the write flow using the industry-standard https://microservices.io/patterns/data/transactional-outbox.html (used by  Netflix, Uber, DoorDash):                                                                                                                            
                                                                                                                                                                                                                                                                               
  Reverse the write flow using the industry-standard Transactional Outbox pattern:                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                    
  1. Atomic DB Write: Chunk metadata + parquet bytes are inserted into PostgreSQL in a single transaction                                                                                                                                                                           
  2. Background Worker: OutboxWorker polls the outbox table and uploads pending entries to S3                                                                                                                                                                                       
  3. Cleanup: Processed entries are periodically cleaned up             

Configuration                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                    
  The outbox worker behavior can be tuned via environment variables defined in src/params.rs. These should be set before starting mosaicod:                                                                                                                                         
                                                                                                                                                                                                                                                                                    
  - MOSAICO_OUTBOX_POLL_INTERVAL_MS (default: 100) - How often the worker checks for pending uploads, in milliseconds                                                                                                                                                               
  - MOSAICO_OUTBOX_BATCH_SIZE (default: 50) - Maximum number of entries processed per polling cycle                                                                                                                                                                                 
  - MOSAICO_OUTBOX_MAX_RETRIES (default: 5) - How many times a failed upload is retried before being abandoned      

 Test
  8 new edge-case unit tests for outbox operations

### DIFF
--- a/mosaicod/Cargo.lock
+++ b/mosaicod/Cargo.lock
@@ -3715,6 +3715,7 @@ checksum = "ee6798b1838b6a0f69c007c133b8df5866302197e404e8b6ee8ed3e3a5e68dc6"
 dependencies = [
  "base64",
  "bytes",
+ "chrono",
  "crc",
  "crossbeam-queue",
  "either",
@@ -3791,6 +3792,7 @@ dependencies = [
  "bitflags",
  "byteorder",
  "bytes",
+ "chrono",
  "crc",
  "digest",
  "dotenvy",
@@ -3833,6 +3835,7 @@ dependencies = [
  "base64",
  "bitflags",
  "byteorder",
+ "chrono",
  "crc",
  "dotenvy",
  "etcetera",
@@ -3868,6 +3871,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2d12fe70b2c1b4401038055f90f151b78208de1f9f89a7dbfd41587a10c3eea"
 dependencies = [
  "atoi",
+ "chrono",
  "flume",
  "futures-channel",
  "futures-core",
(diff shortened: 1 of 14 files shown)


cabba:
Thank you, it will take me a bit of time to do a review since it comes with substantial changes. I also want to fully understand whether the complexity introduced by this pattern is really worth it.

cabba:
I will close this PR since we have changed the way we upload data. See #338

PR #185, opened 2026-02-24
mosaicod-repo split into mosaicod-db and mosaicod-facade

DevMattG:


mosaicod/crates/mosaicod-db/src/sql/mod.rs
```diff
@@ -0,0 +1,6 @@
+pub mod schema;
+pub use schema::*;
+
+pub mod pg_queries;
```
cabba, line 5: I will add `#[cfg(feature="postgres")]` also here. To avoid compiling `pg_queries` module if not required.

PR #186, opened 2026-02-24
Fixed messages payloads in mosaicod 

cabba:
Improved clarity in payloads.

### DIFF
--- a/mosaicod/crates/mosaicod-marshal/src/actions/requests.rs
+++ b/mosaicod/crates/mosaicod-marshal/src/actions/requests.rs
@@ -12,7 +12,7 @@ pub struct Empty {}
 /// Specialized message used to create a new sequence in the platform
 #[derive(Deserialize, Debug)]
 pub struct SequenceCreate {
-    pub name: String,
+    pub locator: String,
     user_metadata: serde_json::Value,
 }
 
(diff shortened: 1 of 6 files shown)


mosaicod/crates/mosaicod-marshal/src/actions/requests.rs
```diff
@@ -50,18 +50,18 @@ impl TopicCreate {
...
 /// Request used to locate a resource deterministically,
 /// by combining the resource name and a unique key.
 /// Used for topics, sequences, or other uuid resources.
-#[derive(Deserialize, Debug)]
-pub struct UploadToken {
-    pub name: String,
-    pub uuid: String,
-}
-
+// #[derive(Deserialize, Debug)]
```
DevMattG, line 59: Better to remove commented code

mosaicod/tests/src/actions.rs
```diff
@@ -150,7 +150,7 @@ pub async fn topic_create(
         body: format!(
             r#"
         {{
-            "name": "{name}",
+            "locator": "{name}",
```
DevMattG, line 153: Shouldn't be better to use also {locator} instead of {name}?

PR #195, opened 2026-02-25
Fixed `sequence_delete` action

### LINKED ISSUE #192: Foreign Key Constraint Violation When Executing SEQUENCE_DELETE Action
When sending the `do_action: SEQUENCE_DELETE` request, the backend returns a database error indicating a foreign key constraint violation. Specifically, the delete operation on table `sequence_t` fails because related records still exist in `session_t`, violating the constraint `session_t_sequence_id_fkey`.

**Error Message:**

```
facade error :: database error :: backend error :: 
error returned from database: 
update or delete on table "sequence_t" violates foreign key constraint 
"session_t_sequence_id_fkey" on table "session_t"
```

**Impact:**
The system does not allow deletion of a sequence if there are associated session records, and the error is not handled gracefully at the application level.

**Likely Cause:**
The backend attempts to delete a sequence without first deleting or handling dependent rows in `session_t`, and no cascading delete is configured for the foreign key relationship.

cabba:
Closes #192

### DIFF
--- a/mosaicod/tests/tests/resources.rs
+++ b/mosaicod/tests/tests/resources.rs
@@ -161,8 +161,11 @@ async fn session_abort(pool: sqlx::Pool<db::DatabaseType>) {
 async fn sequence_delete(pool: sqlx::Pool<db::DatabaseType>) {
     let port = common::random_port();
 
-    let server = common::Server::new(common::HOST, port, pool).await;
-    let mut client = common::Client::new(common::HOST, port).await;
+    let server = common::ServerBuilder::new(common::HOST, port, pool)
+        .build()
+        .await;
+
+    let mut client = common::ClientBuilder::new(common::HOST, port).build().await;
 
     let sequence_name = "test_sequence";
 


cabba:
I need to update cached queries in `.sqlx`

mosaicod/crates/mosaicod-facade/src/session.rs
```diff
@@ -147,6 +143,10 @@ impl Session {
                     .notify(types::NotifyType::Error, msg.clone())
                     .await?,
             );
+        } else {
+            // This is done as last operation, otherwise multiple cals to this function will fail
```
DevMattG: typo: cals -> calls

mosaicod/crates/mosaicod-facade/src/topic.rs
```diff
@@ -323,10 +317,12 @@ impl Topic {
     }
 
     /// Permanently deletes a topic and all its data, be caution
+    ///
+    /// A [`types::DataLossToken`] is required since this call will lead to data losess.
```
DevMattG: typo: losess?

mosaicod/tests/src/actions.rs
```diff
@@ -45,6 +45,34 @@ pub async fn sequence_create(
...
+            locator,
+        )
+        .into(),
+    };
+
+    dbg!(&action);
+
+    let mut stream = client.do_action(action).await.unwrap().into_inner();
+
+    while let Some(result) = stream.message().await.expect("Problem while streaming") {
```
DevMattG, line 66: Is a while loop really needed here?

cabba: Right now, I think removing the while will lead to the same behavior. However, I prefer to reach the end of the stream.

PR #215, opened 2026-02-27
Issue/199/0.3.0

LuigiTer:
Enabled TLS handling to give the user more flexibility. The certificate
path can now be provided directly via `connect()`, or resolved
automatically from the `MOSAICO_TLS_CERT_FILE` environment variable
using `from_env()`.

### DIFF
--- a/scripts/tests.sh
+++ b/scripts/tests.sh
@@ -1,14 +1,5 @@
 #!/usr/bin/env bash
 
-# Mosaico Test Runner
-#
-# Usage:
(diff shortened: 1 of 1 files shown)


scripts/tests.sh
```diff
@@ -234,6 +277,11 @@ main() {
                 run_selected=true
                 shift
                 ;;
+            --integration_with_tls)
```
cabba, line 273: We also need to add this option in the help description

https://github.com/mosaico-labs/mosaico/blob/2a5c8f77030ac81aef18460bec2810b50de8f912/scripts/tests.sh#L1-L11

PR #224, opened 2026-03-02
Added Version endpoint

DevMattG:


mosaicod/crates/mosaicod-core/src/params.rs
```diff
@@ -169,3 +169,12 @@ impl std::fmt::Display for Hidden {
         write!(f, "**********")
     }
 }
+
+/// Returns mosaicod version.
+pub fn get_version() -> String {
```
cabba: i will rename this `version`

cabba (approved):

PR #237, opened 2026-03-03
ci: add ruff lint and format check job for mosaico-sdk-py

LuigiTer:
- Add python-sdk-linter job running ruff format --check and ruff check against mosaico-sdk-py/
- Add python-sdk-linter to integration-tests needs to block on lint failure
- Add --integration_with_tls run in the integration tests step, executed after --integration

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,23 +6,23 @@ on:
       - "main"
       - "release/*"
     paths:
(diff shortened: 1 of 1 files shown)


fdicorato (approved):

cabba (approved):
If we move to Python 3.12 I will change the `python-version` to `3.12`

PR #281, opened 2026-03-17
Resolved datafusion cache problems after session abort

### LINKED ISSUE #193: Corruption Error After Interrupted Run and Sequence Delete (Compressed Image Topics Only)
When a run is interrupted with `Ctrl+C` (triggering `SEQUENCE_DELETE`) and then restarted without restarting the container, subsequent processing fails with an Apache DataFusion Parquet error — but only for rosbag topics using compressed image types.

### Procedure to Reproduce
1. Run:
   ```bash
   # from mosaico root 
   cd docker/quick_start
   docker compose up

   # open another terminal
   # from mosaico-sdk-py root
   cd src/examples
   poetry run python -m ros_injection.main
   ```
2. Press `Ctrl+C` after a while (this triggers sequence deletion)
5. Run (again):
   ```bash
   poetry run python -m ros_injection.main
   ```
5. Let the run complete...

**Result:**
For rosbag topics with compressed image types only, the following error occurs:

```
facade error :: query error :: datafusion backend error :: 
Parquet error: Arrow: Parquet argument error: 
Parquet error: Expected to read 17 values from miniblock got 15
```
### Observed Behavior

* The error appears only for compressed image topics.
* Non-compressed topics are not affected.
* The issue occurs only when the container is not restarted after interruption.
* Suggests partial/corrupted Parquet state or incomplete cleanup after sequence deletion.

### Procedure to Fix / Workaround

1. Run this:
   ```bash
    # from mosaico root 
   cd docker/quick_start
   docker compose up

   # open another terminal
   # from mosaico-sdk-py root
   cd src/examples
   poetry run python -m ros_injection.main
   ```
2. Press `Ctrl+C` after a while (this triggers sequence deletion)
3. **Stop the container** (do not delete the image or remove the volumes)
4. Start container again
5. Run again:   
   ```bash
   poetry run python -m ros_injection.main
   ```
5. Let the run complete...

**Result:**
Everything works correctly; no DataFusion / Parquet errors occur.

### Suspected Cause

An interrupted run leaves the Parquet files or internal state in an inconsistent or partially written state. The `SEQUENCE_DELETE` cleanup appears insufficient when the container remains running, particularly for compressed image topic data.

### Expected Behavior

* Interrupted runs should not corrupt Parquet state.
* `SEQUENCE_DELETE` should fully clean up partial writes.
* Restarting the container should not be required for correct execution.

cabba:
Closes #193 introducing a new data folder (e.g. `data:36f7e`) inside topic, this folder will contain all data chunks for a given topic. The data folder name is computed from the first 5 digits of the topic uuid.

This is done to prevent datafusion caching errors (datafusion was searching old files) when aborting sessions.

### DIFF
--- a/mosaicod/crates/mosaicod-core/src/types/uuid.rs
+++ b/mosaicod/crates/mosaicod-core/src/types/uuid.rs
@@ -33,8 +33,8 @@ impl Uuid {
         !(self.0.is_nil() || self.0.is_max())
     }
 
-    /// Return a non hypened string
-    pub fn non_hypened_string(&self) -> String {
+    /// Return a non hyphened string
+    pub fn non_hyphened_string(&self) -> String {
         self.0.simple().to_string()
     }
 }
--- a/mosaicod/crates/mosaicod-facade/src/topic.rs
+++ b/mosaicod/crates/mosaicod-facade/src/topic.rs
@@ -266,7 +266,6 @@ impl Topic {
             }
         };
 
-        // Cloning  since needs to be moved in side the file format callback
         let data_folder = self.locator.path_data_folder(self.uuid());
 
         let cw = rw::ChunkedWriter::new(self.store.clone(), format, move |chunk_number| {
--- a/mosaicod/crates/mosaicod-core/src/types/resources.rs
+++ b/mosaicod/crates/mosaicod-core/src/types/resources.rs
@@ -161,7 +161,7 @@ impl TopicResourceLocator {
     /// sequence/my/topic/2sr5g
     /// ```
     pub fn path_data_folder(&self, uuid: &Uuid) -> path::PathBuf {
-        let cropped_uuid: String = uuid.non_hypened_string().chars().take(5).collect();
+        let cropped_uuid: String = uuid.non_hyphened_string().chars().take(5).collect();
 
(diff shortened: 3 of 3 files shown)


mosaicod/crates/mosaicod-core/src/types/uuid.rs
```diff
@@ -32,6 +32,11 @@ impl Uuid {
     pub fn is_valid(&self) -> bool {
         !(self.0.is_nil() || self.0.is_max())
     }
+
+    /// Return a non hypened string
+    pub fn non_hypened_string(&self) -> String {
```
DevMattG: typo: hypened -> hyphened

mosaicod/crates/mosaicod-facade/src/topic.rs
```diff
@@ -236,15 +266,15 @@ impl Topic {
...
-        let cw = rw::ChunkedWriter::new(
-            self.store.clone(),
-            self.path(),
-            format,
-            |path, format, idx| {
-                types::TopicResourceLocator::from(path)
-                    .path_data(idx, format.to_properties().as_ref())
-            },
-        )
+        // Cloning  since needs to be moved in side the file format callback
```
DevMattG: Fix this comment

PR #302, opened 2026-03-24
Hotfix 0.2.6

### LINKED ISSUE #301: Out Of Memory problem when running queries on large datasets
User reported OOM problems when running queries on large datasets (see #299 for details).

In our current implementation datafusion query engine uses no disk manager and has no memory pool associated. We need to configure the runtime in order to set both memory pool and disk manager.

Also we need to limit the maximum allowed batch size since running sorting queries (like  the one used when sorting timestamps) can increase exponentially the amount of ram used by the query engine.

cabba:
Closes #301

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,33 +1,30 @@
 # Contributing to Mosaico
 
-Thank you for your interest in contributing to Mosaico!
-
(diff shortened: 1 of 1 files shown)


cabba:
> Good job!

Thank you @fdicorato, very appreciated

PR #355, opened 2026-04-07
refactor!: sequence, session and topic facade refactoring

### LINKED ISSUE #340: Facade refactoring
The current implementation for Sequence, Session and Topic facade structs can be partially misleading. To better indicate that these objects do not maintain an internal state and that having a valid facade object does not necessary mean that the corresponding record on DB still exists we want to perform the following changes:

* Move endpoint::Context struct from mosaicod-server into mosaicod-facade crate 
* Sequence, Session, Topic and Auth structs become opaque handles where all their fields are private
* Associated functions or methods turn into free-functions taking the corresponding handle and the facade::Context as input when necessary

DevMattG:
Closes #340

### DIFF
--- a/mosaicod/crates/mosaicod-facade/src/sequence.rs
+++ b/mosaicod/crates/mosaicod-facade/src/sequence.rs
@@ -22,8 +22,8 @@ impl Handle {
     /// Try to obtain a handle from a sequence locator.
     /// Returns an error if the sequence does not exist.
     pub async fn try_from_locator(
-        locator: types::SequenceResourceLocator,
         context: &Context,
+        locator: types::SequenceResourceLocator,
     ) -> Result<Handle, Error> {
         let mut cx = context.db.connection();
 
@@ -37,7 +37,7 @@ impl Handle {
 
     /// Try to obtain a handle from a sequence UUID.
     /// Returns an error if the sequence does not exist.
-    pub async fn try_from_uuid(uuid: &types::Uuid, context: &Context) -> Result<Handle, Error> {
+    pub async fn try_from_uuid(context: &Context, uuid: &types::Uuid) -> Result<Handle, Error> {
         let mut cx = context.db.connection();
 
         let db_sequence = db::sequence_find_by_uuid(&mut cx, uuid).await?;
@@ -69,9 +69,9 @@ impl Handle {
 /// If a record with the same locator already exists, the operation fails and
 /// the database transaction is rolled back, restoring the previous state.
 pub async fn try_create(
+    context: &Context,
     locator: types::SequenceResourceLocator,
     metadata: Option<SequenceMetadata>,
-    context: &Context,
 ) -> Result<Handle, Error> {
     let mut tx = context.db.transaction().await?;
 
@@ -84,7 +84,7 @@ pub async fn try_create(
     let record = db::sequence_create(&mut tx, &record).await?;
 
     if let Some(mdata) = metadata {
-        metadata_write_to_store(&locator, mdata, context).await?;
+        metadata_write_to_store(context, &locator, mdata).await?;
     }
 
     tx.commit().await?;
@@ -113,9 +113,9 @@ pub async fn all(context: &Context) -> Result<Vec<Handle>, Error> {
 }
 
 async fn metadata_write_to_store(
+    context: &Context,
     locator: &types::SequenceResourceLocator,
     metadata: SequenceMetadata,
-    context: &Context,
 ) -> Result<(), Error> {
     let path = locator.path_metadata();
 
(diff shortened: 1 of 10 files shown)


mosaicod/crates/mosaicod-facade/src/sequence.rs
```diff
@@ -1,266 +1,292 @@
...
+        identifiers: record.into(),
+    })
+}
 
-        tx.commit().await?;
+/// Retrieves all sequences from the database.
+///
+/// Returns a list of all available sequences as [`SequenceResourceLocator`] objects.
+/// This is primarily used for catalog discovery operations.
+pub async fn all(db: db::Database) -> Result<Vec<types::SequenceResourceLocator>, Error> {
```
cabba: Maybe `try_all` since it's returning a `Result`?

mosaicod/crates/mosaicod-facade/src/sequence.rs
```diff
@@ -1,266 +1,292 @@
...
+    let mut cx = db.connection();
+    let records = db::sequence_find_all(&mut cx).await?;
+
+    Ok(records
+        .into_iter()
+        .map(|record| types::SequenceResourceLocator::from(record.locator_name))
+        .collect())
```
cabba: We can return a list of handles, since `db::sequence_find_all` returns a list of records.

To retrieve the `SequenceResourceLocator` from a `Handler` we can add a dedicated function `facade::sequence::locator(handler, ctx)`.

mosaicod/crates/mosaicod-facade/src/sequence.rs
```diff
@@ -1,266 +1,292 @@
...
-    pub async fn session_list(&self) -> Result<Vec<types::Uuid>, Error> {
-        let mut cx = self.db.connection();
+    Ok(topics)
+}
 
-        let sessions = db::sequence_find_all_sessions(&mut cx, &self.locator).await?;
+/// Returns the session list associated with this sequence as vector of session UUIDs
+pub async fn session_list(handle: &Handle, context: &Context) -> Result<Vec<types::Uuid>, Error> {
+    let mut cx = context.db.connection();
+    let sessions = db::sequence_find_all_sessions(&mut cx, &handle.locator).await?;
```
cabba: Not directly related to this PR, but `db::sequence_find_all_sessions` needs to be renamed `db::sequence_find_all_sessions_uuid` for consistency (see line 195 for reference).

mosaicod/crates/mosaicod-facade/src/topic.rs
```diff
@@ -17,442 +16,466 @@ use std::sync::Arc;
...
-    /// Return the inner locator and consumes the facade
-    pub fn into_inner(self) -> types::TopicResourceLocator {
-        self.locator
+/// Creates a new database entry for this topic.
+///
+/// If a record with the same name already exists an error [`Error::TopicAlreadyExists`] is returned.
+///
+/// Additional checks about the scope of the topic are performed. If the topic locator is
+/// not a child of the related sequence locator an error [`Error::Unauthorized`] is returned.
+pub async fn create(
```
cabba: Why `create`? Since in sequence module we have `facade::sequence::try_create`.

mosaicod/crates/mosaicod-facade/src/topic.rs
```diff
@@ -17,442 +16,466 @@ use std::sync::Arc;
...
+    locator: &types::TopicResourceLocator,
```
cabba: locator is cloned inside the function (line 115), avoid borrowing, prefer move.

mosaicod/crates/mosaicod-facade/src/session.rs
```diff
@@ -6,244 +6,273 @@
...
+    pub(super) fn id(&self) -> i32 {
+        self.identifiers.id
+    }
+}
 
-        let manifest =
-            types::SessionManifest::new(self.uuid.clone(), db_session.creation_timestamp());
-        self.manifest_write_to_store(manifest).await?;
+pub async fn try_create(
+    sequence_locator: &types::SequenceResourceLocator,
```
cabba: Also here, locator is cloned (line 69), prefer move.

cabba (changes requested):
I will move `facade::Context` to the beginning of function signatures (it is the same whether it is before or after `Handler`). 

I think this will improve clarity.

mosaicod/crates/mosaicod-facade/src/sequence.rs
DevMattG: Better to leave it as it is

mosaicod/crates/mosaicod-facade/src/sequence.rs
DevMattG: Better to rename it db::sequence_find_all_sessions and to return a vector of session handles

PR #358, opened 2026-04-08
Unify PyArrow Struct and Class Attributes Declaration 

LuigiTer:
- Add `MosaicoType` class with `Annotated` aliases for all scalar PyArrow  types (int8/16/32/64, uint8/16/32/64, float16/32/64, bool, string, large_string, binary, large_binary) and a `list_()` static method for variable-length and fixed-size list columns
- Add `MosaicoField` factory returning a native `pydantic.FieldInfo` 
- Refactor `Serializable.__pydantic_init_subclass__` to derive `__msco_pyarrow_struct__` automatically by introspecting `model_fields` and extracting the embedded `pa.DataType` from each `Annotated` annotation
- Add unit tests covering scalar types, list types, nullable structs, custom `Annotated` fallback, and mixin field aggregation.
- Update documentation with full API reference, usage examples, and mixin composition guide.


doc/docs/SDK/ontology.md
```diff
@@ -87,6 +87,133 @@ The following sensors are supported as first-class data types in this module:
 !!! warning "Experimental Ontologies"
     The `futures` module is a transitional area where ontologies under active experimentation are hosted before graduating to the stable ontology set. Field definitions, unit conventions, and structural relationships are **not yet considered final** and will be refined based on feedback from real-world integrations and adopters. Once an ontology reaches sufficient maturity and coverage, it will be promoted out of `futures` and into the core, production-ready modules.
 
+## MosaicoType & MosaicoField
+When defining ontologies in the Mosaico SDK, every class attribute carries two pieces of information:
+
+* **Python type** used by Pydantic for validatio and IDE support.
```
fdicorato: validatio

doc/docs/SDK/ontology.md
```diff
@@ -87,6 +87,133 @@ The following sensors are supported as first-class data types in this module:
...
+ 
+    # Fixed-size list of 3 float32 (e.g. an RGB vector)
+    color: MosaicoType.list_(MosaicoType.float32, list_size=3)
+ 
+    # Works with raw Python primitives too
+    tags: MosaicoType.list_(str)
+
+    #Works with other Pydantic models with pyarrow struct
+    vec: MosaicoType.list_(Vector3d)
+```
```
fdicorato, line 192: Please, add here a brief discussion about what happens if the user uses the standard python `list/List` types, instead of  `MosaicoType.list_`.
* Is it allowed?
* What is the difference (under the hood) between `tags: MosaicoType.list_(str)` and `tags: list[str]`?
* When the `list_()` method should be preferred?
* When choosing between `list_()` and `list` is irrelevant?

doc/docs/SDK/ontology.md
```diff
@@ -87,6 +87,133 @@ The following sensors are supported as first-class data types in this module:
...
+ 
+#### Custom Arrow types
+ 
+For specialised Arrow types not covered by the built-in aliases, you can always use a raw `Annotated` annotation embedding the PyArrow type directly. The schema builder resolves the embedded metadata transparently, exactly like a `MosaicoType` alias:
+
+```python
+from mosaicolabs import Serializable, MosaicoField
+class MyOntology(Serializable):
+    ts: Annotated[int, pa.timestamp("us", tz="UTC")] = MosaicoField(description="UTC timestamp.")
```
fdicorato: Proposal (not a requirement): we can hide the `Annotated` construct in this case also, behind a static method like (choose another name if you don't like):
```python
class MosaicoType:
   #... the current impl here

   @staticmethod
   def make(py_type: Type, pa_type: pa.DataType)->Annotated: # <- add this
        return Annotated[py_type, pa_type]
```

What do you think?

doc/docs/SDK/ontology.md
```diff
@@ -87,6 +87,133 @@ The following sensors are supported as first-class data types in this module:
...
+    x:     MosaicoType.float32 = MosaicoField(description="X coordinate")
+    y:     MosaicoType.float32 = MosaicoField(description="Y coordinate")
+    label: Optional[MosaicoType.string] = MosaicoField(
+        default=None, description="Point label")
+    score: MosaicoType.float32 = MosaicoField(nullable=True)
+```
+
+With `MosaicoField` you can define the `default` value of your attribute, the `nullable` attribute of pyarrow field and also a `description`.
+
+### Nullability and Parquet V2
```
fdicorato, line 272: please, rephrase this section; in particular this part: '*Therefore all four fields are declared nullable for pyarrow*.' is not logically related to '*a quaternion with missing components is meaningless and cannot be constructed*'. The logical deduction would be that the fields should be defined non-nullable. 

The issue here is that the nullability of the pyarrow fields in the Quaternion class is set to True due to how  ParquetV2 handles `null` optional columns during data reading, in classes like IMU, where the field `orientation` is an `Optional[Quaternion]`. If this column in the parquet file is `null` and the pyarrow fields of (x, y, z, w) are not nullable in the schema, the `orientation = None` would be reconstructed (during read) as: `orientation = Quaternion(x=0, y=0, z=0, w=0)`.

doc/docs/SDK/ontology.md
```diff
@@ -227,6 +353,38 @@ class MySensor(Serializable, VarianceMixin):
...
 
+### Extending with Mixins
+
+One of the most powerful consequences of building on top of Pydantic model fields is how natural **mixin composition** becomes. Because every field, including its Arrow type metadata, lives in `model_fields`, you can split concerns into focused mixin classes and combine them freely without any additional registration or schema merging step.
+
+```python
+from mosaicolabs import MosaicoType, MosaicoField, Serializable
+
+
+class GeometryMixin(Serializable):
```
fdicorato: Are you sure about this definition of the *mixins*? Inheriting from `Serializable`, makes them automatically eligible to be ontology models, i.e. you can construct a mosaico `Message` like this:
```python
msg = Message(
   timestamp_ns = 12345678901234,
   data = GeometryMixin(...),
)
```
But this might not be the case.

If the mixins just provide additional common data (like `CovarianceMixin`, `VarianceMixin`, etc.) it would be sufficient for them to inherit from BaseModel only. Should an ontology model need some of them, it must only inherit from them, before inheriting from `Serializable`, i.e.

```python
class GeometryMixin(BaseModel):
     ....

class MyClassNeedingGeometry(GeometryMixin, Serializable):
     ....
```

doc/docs/SDK/ontology.md
```diff
@@ -335,26 +487,21 @@ This example demonstrates a custom sensor for environmental monitoring that trac
...
-    )
-
-    # --- 2. Define Python Fields (Must match schema exactly) ---
-    temperature: float
-    humidity: Optional[float] = None
-    pressure: Optional[float] = None
+    # --- 1. Define the Wire Schema ---
+    temperature: MosaicoType.float32
+    
+    humidity: Optional[MosaicoType.float32] = MosaicoField(
```
fdicorato, line 604: One question: is there any difference under the hood between these?
`pressure: Optional[MosaicoType.float32] = MosaicoField(default= None, nullable=True)` and `pressure: Optional[MosaicoType.float32] = None`

doc/docs/SDK/ontology.md
```diff
@@ -87,6 +87,133 @@ The following sensors are supported as first-class data types in this module:
...
+from mosaicolabs import MosaicoField, MosaicoType, Serializable
+
+class MyOntology(Serializable):
+    x: MosaicoType.float32
+    y: Optional[MosaicoType.float32] = None
+
+```
+
+In the above example `x` attribute will be converted in a **not nullable** pyarrow field, while `y` attribute will be converted in a **nullable** pyarrow field.
+
```
fdicorato, line 167: I do not see any reference to the fallback types, i.e. what happens if the ontology is defined as:

```python
class MyOntology(Serializable):
    x: float
    y: Optional[float] = None
```

Please, add few lines about this

mosaico-sdk-py/src/mosaicolabs/models/data/geometry.py
```diff
@@ -1263,7 +1192,9 @@ class Transform(
         ```
     """
 
-    target_frame_id: Optional[str] = None
+    target_frame_id: Optional[MosaicoType.string] = MosaicoField(
+        default=None, nullable=True, description="Target frame identifier."
```
fdicorato: is it necessary to set both `default=None` and `nullable=True`? Does it suffice to set `default=None` and deduce `nullable=True`?

mosaico-sdk-py/src/mosaicolabs/models/futures/radar.py
```diff
@@ -46,11 +44,10 @@ class Radar(Serializable):
         ax: X component of the acceleration of each detection in m/s² (optional).
         ay: Y component of the acceleration of each detection in m/s² (optional).
         radial_speed: Radial speed of each detection in m/s (optional).
-        extra_attributes: Additional manufacturer-specific attributes serialised as
-            raw binary data (optional).
+
 
     Note:
-        List-typed fields are **not queryable** via the `.Q` proxy. The `.Q` proxy
+        MosaicoType.list_(-typed fields are **not queryable** via the `.Q` proxy. The `.Q` proxy
```
fdicorato: missing a closing ')'

mosaico-sdk-py/src/mosaicolabs/models/mixins.py
fdicorato, line 1: Please, check the correct inheritance of the mixins in the classes that inherit from them (e.g. `IMU` inherit from `CovarianceMixin`). 
To do so, add unit tests for checking the existence of the pyarrow fields of the mixins in the `__msco_pyarrow_struct__` variable of the ontology models; for example: check that `IMU.__msco_pyarrow_struct__` contains the `covariance` and `covariance_type` pyarrow fields

mosaico-sdk-py/src/mosaicolabs/ros_bridge/data_ontology/point_cloud.py
```diff
@@ -27,22 +28,17 @@ class PointField(Serializable):
...
-    name: str
+    name: MosaicoType.string = MosaicoField(description="Name of the field.")
     """The name of the point field."""
-    offset: int
+    offset: MosaicoType.uint32 = MosaicoField(
+        description="Staring position of the field."
+    )
     """The Offset from start of point struct."""
-    datatype: PointFieldDataType
+    datatype: PointFieldDataType = MosaicoField(description="Datatype of the field.")
```
fdicorato, line 37: How this enum is mapped into pyarrow schema?

fdicorato (changes requested):

mosaico-sdk-py/src/mosaicolabs/ros_bridge/data_ontology/point_cloud.py
LuigiTer: There is the specific Resolver that resolve the base type of the Enum

mosaico-sdk-py/src/mosaicolabs/models/mixins.py
LuigiTer: The present tests cover this particular aspect.

mosaico-sdk-py/src/mosaicolabs/models/data/geometry.py
LuigiTer: In the case of default=... but nullable=True, it can create confusion. I can add the specific check in the factory method.

fdicorato (changes requested):
Please update the section `#### Custom Arrow types` in file `doc/docs/SDK/ontology.md`:
* change `ts: Annotated[int, pa.timestamp("us", tz="UTC")]` to `ts: MosaicoType.annotate(int, pa.timestamp(...))`
* briefly describe what `MosaicoType.annotate()` is and how it is used

fdicorato (approved):
Great job! thank you

PR #365, opened 2026-04-09
refactor(mosaicod)!: session manifest removed from Store

### LINKED ISSUE #341: Remove session manifest file from Store
Remove the session-\<uuid\>.json file from the Store (no data for a session should be in the Store). A SessionManifest object can still be created from the corresponding session record in the database.

DevMattG:
BREAKING CHANGES: GetFlightInfo endpoint now returns the list of topics even for open sessions (not locked)

Closes #341


mosaicod/crates/mosaicod-db/src/sql/pg_queries/session_record.rs
```diff
@@ -109,3 +121,27 @@ pub async fn session_find_all_topics(
     .fetch_all(exe.as_exec())
     .await?)
 }
+
+pub async fn session_update_completion_ts(
```
cabba: Maybe `session_update_completion_tstamp`? To avoid adding another shorthand (`ts`)  for the timestamp.

cabba (approved):

PR #374, opened 2026-04-10
refactor(mosaicod): Topic FlightInfo app_metadata generated only with info from DB

### LINKED ISSUE #344: Get topic info only from DB to create GetFlightInfo appmetadata
At the moment to create the topic appmetadata in a GetFlightInfo response, data from both Store and DB is used. We want to remove this dependency from the Store in favor of the database. To perform this the only field missing inside the topic DB table is the "completion_unix_tstamp" (completion timestamp for the topic data load).

TopicManifest type will also be renamed to TopicMetadata, and manifest.json will be renamed to metadata.json. "locked" entry will be removed ("completed_at" is sufficient to determine when a topic's data load finished)

DevMattG:
Closes #344


mosaicod/crates/mosaicod-db/src/sql/pg_queries/topic_record.rs
```diff
@@ -333,3 +374,15 @@ pub async fn topic_from_query_filter(
     trace!("query returned {} results", r.len());
     r.into_iter().collect()
 }
+
+/// Returns true if the topic is locked.
+pub async fn topic_locked(exe: &mut impl AsExec, topic_id: i32) -> Result<bool, Error> {
```
cabba: We want to continue calling this `locked`?

mosaicod/crates/mosaicod-facade/src/topic.rs
```diff
@@ -145,62 +146,79 @@ pub async fn try_create(
...
 
+pub async fn archived(context: &Context, handle: &Handle) -> Result<bool, Error> {
+    let mut cx = context.db.connection();
+    Ok(db::topic_locked(&mut cx, handle.id()).await?)
+}
+
 /// Finalize the write procedure of the topic. The topic is locked and additional data are
-/// consolidated (e.g. manifest, timestamp bounds). This function is intended to be called by
+/// consolidated (e.g. metadata, timestamp bounds). This function is intended to be called by
 /// [`TopicWriterGuard`] to finalize the writing process.
```
cabba: Please update the reference to `TopicWriter`

cabba (approved):

mosaicod/crates/mosaicod-db/src/sql/pg_queries/topic_record.rs
DevMattG: Ok, we'll rename it later

cabba (approved):

PR #383, opened 2026-04-13
refactor(mosaicod): Sequence FlightInfo app_metadata generated only with info from DB

### LINKED ISSUE #373: Get sequence info only from DB to create GetFlightInfo appmetadata
At the moment to create the sequence appmetadata in a GetFlightInfo response, data from both Store and DB is used. We want to remove this dependency from the Store in favor of the database.

- [x]  rename SequenceManifest type to SequenceMetadata

- [x] rename manifest.json to metadata.json

DevMattG:
Closes #373

### DIFF
--- a/mosaicod/crates/mosaicod-facade/Cargo.toml
+++ b/mosaicod/crates/mosaicod-facade/Cargo.toml
@@ -28,3 +28,4 @@ tokio = { workspace = true }
 sqlx = { workspace = true }
 mosaicod-store = { workspace = true, features = ["testing"] }
 mosaicod-db = { workspace = true, features = ["postgres", "testing"] }
+serde_json = { workspace = true }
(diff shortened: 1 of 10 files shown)


mosaicod/crates/mosaicod-facade/Cargo.toml
```diff
@@ -28,3 +28,4 @@ tokio = { workspace = true }
 sqlx = { workspace = true }
 mosaicod-store = { workspace = true, features = ["testing"] }
 mosaicod-db = { workspace = true, features = ["postgres", "testing"] }
+serde_json = { workspace = true }
```
cabba, line 31: Dev dependency?

cabba (changes requested):

mosaicod/crates/mosaicod-facade/Cargo.toml
DevMattG: Without it tests inside facade::sequence.rs don't compile

cabba (approved):

PR #433, opened 2026-04-22
refactor(mosaicod)!: changed create, finalize, do_put and delete actions logic for sequence, session and topic

### LINKED ISSUE #345: Refactor create, finalize, do_put and delete actions for sequence, session and topic
Considering the DB as the only origin of truth and the decoupling between the resource locator and the physical storage it is possible to simplify the implementation logic for the following actions (making them more robust at the same time): create, finalize, do_put, and delete.

### Create

Step | Component | Description
-- | -- | --
1 | STORE | A folder with a random name for the sequence and the metadata.json are created
2 | DB | A new “sequence” entry is created and every field is filled. “path_in_store” contains the random name of the folder generated at the previous step

__Note__: with this configuration it is possible for two concurrent create to accomplish step 1, generating two different folders in the Store that refer to the same sequence (same locator name). This is safe thanks to step 2, where only one action wins creating the entry on the DB, while the other fails leaving an orphan folder in the Store (not referenced by a DB record). It will be up to the cleaning routine to delete it later from the Store.

### Delete

Step | Component | Description
-- | -- | --
1 | DB | The “sequence” record is removed.

__Note__: ON DELETE CASCADE constraint must be applied to both session and topic sequence_id foreign keys. 

### Create

Step | Component | Description
-- | -- | --
1 | DB | A new “session” entry is created and every field is filled except for “completion_unix_ts”

### Finalize

Step | Component | Description
-- | -- | --
1 | DB | If the session contains at least one topic and every associated topic is finalized, then the “session” record can be updated with the following data:completion_unix_ts. Otherwise, an error is returned

__Note__: two different cases of failure are possible:
* an error occurs while a single finalize tries to access the DB or update the record
* two concurrent finalize try to update the record and the second returns an error because the completion_unix_ts has already been set

In both cases the action returns an error leaving the session open. It’s up to the user to decide what to do with it.

### Delete

Step | Component | Description
-- | -- | --
1 | DB | “Session” record is removed.

__Note__:
* ON DELETE CASCADE constraint must be applied to topic's session_id foreign key.
* this action is sufficient for the background cleaning routine will take charge of marking all associated folders in the Store as TO_DELETE (creating a TO_DELETE file). 

### Create

Step | Component | Description
-- | -- | --
1 | DB | A new “topic” entry is created and every field is filled except for: completion_unix_ts, chunks_number, total_bytes, start_index_timestamp, end_index_timestamp, path_in_store

### Do_Put

Step | Component | Description
-- | -- | --
1 | STORE | A folder with a random name for the topic containing the metadata.json is created on the Store. At this point metadata.json is missing only the “completed_at” key.
2 | DB | Topic record’s “path_in_store” is set to the random folder path created at the previous step if it’s not already set.

__Note__: in case of concurrent do_put, at step 1 a random folder is created for each action, but only the first action entering step 2 and setting `path_in_store` can continue, causing the others to return an error. Stale folders left from step 1 are then removed by the cleaning routine.

### Finalize

Step | Component | Description
-- | -- | --
1 | STORE | “completed_at” is added to metadata.json
2 | DB | “Topic” record is updated with the following data: completion_unix_ts, chunks_number, total_bytes, start_index_timestamp, end_index_timestamp

__Note__: in case of failure for step 1 or 2, finalize returns an error leaving the topic as it is. It’s up to the user to decide what to do next.

### Delete

Step | Component | Description
-- | -- | --
1 | DB | “Topic” record is removed

Note: this action is sufficient for the background cleaning routine will take charge of marking the associated folder in the Store as TO_DELETE.

DevMattG:
Closes #345 

BREAKING CHANGES:
* added ON DELETE CASCADE to session and topic foreign keys, database migration or reset required.


DevMattG:
@cabba pay attention to the change I made to facade::topic::arrow_schema()

mosaicod/crates/mosaicod-db/src/sql/pg_queries/topic_record.rs
```diff
@@ -99,18 +99,18 @@ pub async fn topic_find_all(exe: &mut impl AsExec) -> Result<Vec<schema::TopicRe
...
 /// elsewhere. Improper use can lead to data inconsistency or loss.
 pub async fn topic_delete(
     exe: &mut impl AsExec,
-    loc: &types::TopicLocator,
+    topic_id: i32,
     _: types::DataLossToken,
 ) -> Result<(), Error> {
-    warn!("(data loss) deleting topic `{}`", loc);
-    sqlx::query!("DELETE FROM topic_t WHERE locator_name=$1", loc as &str)
+    warn!("(data loss) deleting topic with id={} from DB", topic_id);
```
cabba:
```suggestion
    warn!("(data loss) deleting topic record with id={}", topic_id);
```

LuigiTer (approved):
The testing part is ok for me

cabba (changes requested):

cabba (approved):

PR #438, opened 2026-04-23
feat(mosaicod): add support for list of fingerprints in revoke

### LINKED ISSUE #425: Expand mosaicod api-key revoke to get a list of fingerprint
We need to enhance `mosaicod api-key revoke` command to accepts a list of fingerprints instead of only one.

emekaokoli19:
## PR: Extend API key revoke command to support multiple fingerprints

This PR extends the mosaicod CLI API-key revoke command to support revoking multiple API key fingerprints in a single command.

### Changes
- Updated ApiKey::Revoke CLI subcommand to accept Vec<String> instead of a single fingerprint
- Modified revoke handler to iterate over all provided fingerprints
- Each fingerprint is processed independently against the database(I cloned the db)
### Behaviour
- Valid fingerprints are revoked successfully
- Invalid or non-existent fingerprints are reported individually
- One failure does not block revocation of remaining keys
### Test
Setup up postgres and tested locally 
### Closes
Fixes #425


mosaicod/bin/mosaicod/src/command/api_key.rs
```diff
@@ -117,11 +118,20 @@ pub fn auth(auth: ApiKey) -> Result<()> {
...
+        ApiKey::Revoke { fingerprints } => {
             let res: core::error::PublicResult<()> = rt.block_on(async {
-                let fauth = facade::Auth::try_from_fingerprint(&fingerprint, db).await?;
-
-                fauth.delete().await?;
+                for fingerprint in fingerprints {
+                    match facade::Auth::try_from_fingerprint(&fingerprint, db.clone()).await {
+                        Ok(fauth) => {
+                            if let Err(e) = fauth.delete().await {
+                                eprintln!("Failed to revoke {}: {}", fingerprint, e);
```
cabba: I think you can return directly the error provided by the `.delete()` since the facade returns an error that implements a `core::error::PublicResult`. And will be mapped to an internal error.

mosaicod/bin/mosaicod/src/command/api_key.rs
```diff
@@ -117,11 +118,20 @@ pub fn auth(auth: ApiKey) -> Result<()> {
...
+                            }
+                        }
+                        Err(e) => {
+                            eprintln!("Invalid fingerprint {}: {}", fingerprint, e);
```
cabba: I will add a new `core::error::ErrorKind::InvalidFingerprint` and return this one. See line 92 on how to return a `core::Error`.

cabba (changes requested):
Hi @emekaokoli19 thanks for the PR. 

Looks good! You only need to fix a few error return types. 
Usually from CLI commands (but also in the server) we return `core::Error`-s to standardize well the output for end users.

cabba (approved):

PR #439, opened 2026-04-23
feat(mosaicod): Improve existing TLS tests Coverage in mosaicod

### LINKED ISSUE #424: Improve existing TLS tests Coverage in mosaicod
While some tls backend tests already exist, they do not cover important edge cases. 

Current tests do not cover certificate-related failure scenarios. 
The following cases should be added:

- [ ] Missing certificate: connection attempted with no cert provided
- [ ] Wrong certificate: connection attempted with an invalid or mismatched cert

LuigiTer:
Added tls tests for:

- empty tls server/client cert
- wrong tls server/client cert/key

Close https://github.com/mosaico-labs/mosaico/issues/424

### DIFF
--- a/mosaicod/tests/src/common.rs
+++ b/mosaicod/tests/src/common.rs
@@ -148,7 +148,7 @@ impl Server {
         let _ = tokio::join!(self.server_join_handle);
     }
 
-    pub async fn is_crashed(&self) -> bool {
+    pub async fn is_shutdown(&self) -> bool {
         self.server_join_handle.is_finished()
     }
 }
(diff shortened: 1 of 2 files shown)


mosaicod/tests/src/common.rs
```diff
@@ -139,6 +147,10 @@ impl Server {
         self.shutdown.shutdown();
         let _ = tokio::join!(self.server_join_handle);
     }
+
+    pub async fn is_crashed(&self) -> bool {
+        self.server_join_handle.is_finished()
+    }
```
cabba, line 153: I will change this into something like `.is_shutdown()` or `.is_running()` (checking the negative condition). `.is_crashed()` implies that the server terminated for a internal condition, while we can call `.shutdown()` to terminate the server gracefully.

cabba (changes requested):

cabba (approved):
🚀

PR #452, opened 2026-04-24
fix(py): register futures and ROS data_ontology modules in __init__.py

### LINKED ISSUE #450: Add missing imports to package-level `__init__.py` to ensure Serializable registration
Deserialization of messages from certain topics fails because data types from `futures` and ROSBridge specific data ontology models are not registered in the `Serializable` registry at import time. The error occurs when the relevant modules are never imported, so their registration side-effects are not triggered.

### Fix
Explicitly import the necessary modules in the package-level `__init__.py` to force registration during package initialization.

### Changes

Update `mosaicolabs.__init__.py` as follows:

```python
from .models.data import (
    # older imports...

    # Force registration into Serializable registry
    futures as futures,
)

# After .models imports, add:
from .ros_bridge import data_ontology
```

### Notes
* This approach relies on import side-effects; consider documenting this behavior to avoid accidental removal in the future.
* A longer-term solution could involve explicit registration hooks instead of implicit import-based registration.

fdicorato:
This PR fixes the registration errors due to missing global import of `futures` and ROS-related `data_ontology` modules.

**No breaking changes are expected with this fix.**

Closes #450

### DIFF
--- a/mosaico-sdk-py/src/mosaicolabs/__init__.py
+++ b/mosaico-sdk-py/src/mosaicolabs/__init__.py
@@ -52,6 +52,11 @@ from .models import (
     MosaicoType as MosaicoType,
     Serializable as Serializable,
     VarianceMixin as VarianceMixin,
+    # Force registration into Serializable registry
+    # (fix deserializing messages from topics that contain these data types,
+    # with no registration side-effects)
+    # !!DO NOT REMOVE!!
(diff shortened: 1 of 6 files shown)


pucciland95 (approved):

fdicorato:
fix merged from release mosaico-py/v0.4.1 (#465)

PR #455, opened 2026-04-24
feat(mosaicod): optional Arrow IPC body compression on DoGet streams

facontidavide:
Adds MOSAICOD_FLIGHT_IPC_COMPRESSION env var (none/lz4/zstd, default none) that wires arrow::ipc::CompressionType into the FlightDataEncoderBuilder in endpoint::do_get. The encoded IPC bodies are column-aware, complementary to the gRPC-channel GZIP introduced in #361 (the two layers can be combined or used independently).

- Cargo.toml: enable the `ipc_compression` feature on the `arrow` crate so LZ4_FRAME and ZSTD codecs are linked.
- mosaicod-core/params: add `IpcCompression` enum with FromStr parsing and a `flight_ipc_compression: Param<IpcCompression>` field. Invalid values panic at startup (matching the existing Param::optional behavior) so the operator gets immediate feedback rather than a runtime DoGet failure.
- mosaicod-server/endpoint/do_get: build IpcWriteOptions from the param and pass via FlightDataEncoderBuilder::with_options. Default `None` short- circuits to the unmodified default options for full backward compatibility.
- env.devel + CHANGELOG.md: document the new variable.

Backward compatibility: any Arrow Flight client with the matching codec linked decodes transparently (pyarrow >= 12, arrow-cpp >= 12 with LZ4/ZSTD, arrow-rs any 5x with `ipc_compression` enabled). Default `none` means existing deployments see no behavior change unless the operator opts in.

cargo check, cargo clippy --workspace --all-targets -- -D warnings, and cargo fmt --all -- --check all pass clean.

### DIFF
--- a/mosaicod/CHANGELOG.md
+++ b/mosaicod/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Features
+
+- Added optional Arrow IPC body buffer compression on outbound DoGet streams,
+  selected by the `MOSAICOD_FLIGHT_IPC_COMPRESSION` environment variable
+  (`none` default, `lz4`, or `zstd`). Column-aware and complementary to the
+  gRPC-channel GZIP enabled by `--gzip`. Backward compatible: any Arrow Flight
+  client with the matching codec linked decodes transparently.
+
 
 ## [0.3.0] - 2026-30-03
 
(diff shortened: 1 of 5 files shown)


cabba:
Closing this out as we have just merged a global performance update that enables `LZ4_FRAME` compression by default for downloads (#453).

Thanks anyway for the PR 🚀

PR #466, opened 2026-04-28
refactor!: created session locator entity

### LINKED ISSUE #437: Add session locator entity
### Problem

At the moment a GetFlightInfo request for a sequence retrieves the list of all its associated sessions (finalized and not). Within the info about a session the UUID is also provided, letting the door open to unwanted operations. For example a user could connect to a still open session (not created by him) and upload topics.

### Solution

Create a Session locator entity to return instead of the UUID inside GetFlightInfo responses.

To still let a user with the right permissions delete a session (open or finalized), the session delete action must accept the session locator instead of its UUID.

The accepted locator format is \<sequence\>:\<session\>, where session is a ULID generated by the server. No slashes ('/') allowed.

The session create action must then return this newly created locator along with its UUID.

DevMattG:
BREAKING CHANGES:
- session create now returns session locator along with UUID
- session delete requires session locator instead of session UUID
- GetFlightInfo for sequence returns session locators instead of UUIDs

Closes #437


mosaicod/crates/mosaicod-core/src/types/resources.rs
```diff
@@ -51,111 +31,25 @@ impl std::fmt::Display for ResourceKind {
...
-        write!(f, "{}", self.inner)
-    }
-}
+/// Checks if value has symbols not admitted for a locator.
+///
+/// The following criteria must be met:
+/// - non-ASCII chars are not allowed
+/// - special symbols `! " ' * £ $ % &` are not allowed
+fn has_invalid_symbols(value: &str, others: Option<&[char]>) -> bool {
+    let mut invalid_chars = vec!['!', '\"', '\'', '*', '£', '$', '%', '&', '.', ' '];
```
LuigiTer: `invalid_chars` vector is rebuilt on every call. Since the list is fixed, it can be lifted to a `const` to avoid the allocation.

mosaicod/crates/mosaicod-core/src/types/resources.rs
```diff
@@ -181,75 +79,71 @@ impl TopicLocator {
...
+        Some(self.cmp(other))
     }
 }
 
-impl PartialEq<TopicLocator> for &str {
-    fn eq(&self, other: &TopicLocator) -> bool {
-        self == &other.inner
+impl Ord for TopicLocator {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        let ord = self.sequence.cmp(&other.sequence);
```
LuigiTer: you can use `then_with()` to make this less verbose

```suggestion
        self.sequence
            .cmp(&other.sequence)
            .then_with(|| self.name.cmp(&other.name))
```

mosaicod/crates/mosaicod-server/src/endpoint/get_flight_info.rs
LuigiTer: There is a silent error for `data_info` here.

mosaicod/crates/mosaicod-core/src/types/resources.rs
cabba: I think the better approack is to have a `static INVALID_CHAR: &[char] = [ ... ] ` otherwise with `const` we will have runtime allocations.

mosaicod/crates/mosaicod-server/src/endpoint/get_flight_info.rs
```diff
@@ -27,134 +27,128 @@ pub async fn get_flight_info(ctx: &facade::Context, desc: FlightDescriptor) -> R
...
+                );
+
+                let schema = topic_arrow_schema_with_metadata(
+                    metadata.ontology_metadata,
+                    &topic_handle,
+                    ctx,
+                )
+                .await?;
+
+                let flight_info = FlightInfo::new()
```
LuigiTer: The two branches have common logic, you can define a function instead

mosaicod/crates/mosaicod-core/src/types/resources.rs
```diff
@@ -428,79 +322,138 @@ pub struct TopicDataInfo {
...
+    pub sequence: SequenceLocator,
+    /// Name of the session (it does not contain the sequence prefix, but only the session ULID).
+    name: String,
+}
+
+impl SessionLocator {
+    /// Creates a new session locator for the given parent sequence.
+    pub fn new(parent: SequenceLocator) -> Self {
+        Self {
+            name: format!("{}", ulid::Ulid::new()),
```
cabba: If `Ulid` has `.to_string()` I think will be better to use it, more readable.

mosaicod/crates/mosaicod-server/src/endpoint/get_flight_info.rs
LuigiTer, line 26: Consider extracting `handle_cmd` and a helper per locator kind (`handle_sequence`, `handle_topic`) the nested `match` + `if-else` + inner `return` is hard to follow, and each branch would be easier to test in isolation.

LuigiTer (changes requested):

mosaicod/crates/mosaicod-db/src/sql/schema/session_record.rs
```diff
@@ -26,16 +28,28 @@ impl SessionRecord {
...
 
+    /// Returns the resource locator for this session.
+    ///
+    /// Because a [`SessionRecord`] should only be created using [`SessionRecord::new`], that requires a [`types::SessionLocator`],
+    /// we can assume the locator value inside the DB is always valid. It should panic only if somebody
+    /// changed it manually directly inside the database.
+    pub fn locator(&self) -> types::SessionLocator {
+        self.locator_name
+            .parse()
+            .unwrap_or_else(|_| panic!("Invalid session locator in DB {}", self.locator_name))
```
cabba, line 50: Maybe use `expect`?

cabba (changes requested):

mosaicod/crates/mosaicod-server/src/endpoint/get_flight_info.rs
DevMattG: We need to discuss about it and open a separate issue if necessary

mosaicod/crates/mosaicod-db/src/sql/schema/session_record.rs
DevMattG: `.expect`  does not pass clippy check because of the string formatting I should use inside it

LuigiTer (approved):

cabba (approved):

PR #478, opened 2026-05-04
feat(mosaicod): TCP bind tests failures caused by random port collisions

LuigiTer:
Now random port is demanded to OS and protected by guard

### DIFF
--- a/mosaicod/tests/src/common.rs
+++ b/mosaicod/tests/src/common.rs
@@ -1,5 +1,4 @@
 use arrow_flight::flight_service_client::FlightServiceClient;
-use futures::lock::Mutex;
 use mosaicod_core::params;
 use mosaicod_core::types;
 use mosaicod_db as db;
@@ -11,7 +10,10 @@ use serde::Deserialize;
 use std::fs;
 use std::net::TcpListener;
 use std::sync::Arc;
+use std::sync::Mutex;
 use tonic::service::interceptor;
+
+static PORT_LOCK: Mutex<()> = Mutex::new(());
 
 /// The local loopback address for testing.
 pub const HOST: &str = "127.0.0.1";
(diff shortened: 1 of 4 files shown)


mosaicod/tests/src/common.rs
```diff
@@ -8,25 +8,20 @@ use mosaicod_server::{self as server, flight::ShutdownNotifier};
 use mosaicod_store as store;
 use serde::Deserialize;
 use std::fs;
+use std::net::TcpListener;
 use std::sync::Arc;
+use std::sync::Mutex;
 use tonic::service::interceptor;
 
+static PORT_LOCK: Mutex<()> = Mutex::new(());
```
cabba, line 19: Add a comment to document why a mutex is used.

mosaicod/tests/src/common.rs
```diff
@@ -51,6 +45,7 @@ async fn start_server(
     db::testing::Database,
     store::testing::Store,
     query::TimeseriesEngineRef,
+    u16,
```
cabba: Since we are returning a lot of data, I think the best approach is to return a proper struct instead.

mosaicod/tests/src/common.rs
```diff
@@ -65,6 +60,13 @@ async fn start_server(
         .unwrap(),
     );
 
+    let guard = PORT_LOCK.lock();
+
+    let port = {
+        let tcp_listner = TcpListener::bind(format!("{}:0", HOST)).unwrap();
+        tcp_listner.local_addr().unwrap().port()
+    };
```
cabba: Add some docs here to explain why this is done.

cabba (changes requested):

DevMattG (approved):

cabba (approved):

PR #492, opened 2026-05-11
fix(py): Fixes array concatenation correctly in SyncTransformer and add `VideoDecodingTransformer`

fdicorato:


mosaico-sdk-py/src/mosaicolabs/ml/decoding_transformer.py
```diff
@@ -0,0 +1,200 @@
...
+                        format=img_format,
+                        context=topic,  # Isolates decoder state per topic
+                    )
+                # Stateless fallback for JPEGs/PNGs
+                else:
+                    pil_img = self._stateless_codec.decode(
+                        img_bytes=img_bytes, format=img_format
+                    )
+
+                decoded_images.append(pil_img)
```
pucciland95, line 188: decode() may return None. Do we accept None types in decoded_images?

pucciland95 (approved):
Everything seems ok, just a tiny comment

PR #494, opened 2026-05-13
feat(mosaicod): implemented the background cleanup routine

### LINKED ISSUE #343: Implement the background cleanup routine
A cleanup routine is necessary to keep the data in the Store totally aligned with the DB. This service will be implemented as a background thread running for every server instance and will be in charge of deleting the obsolete files (not associated with any entry in the DB) in the Store after a retention period.

To synchronize the cleanup routine between multiple servers and avoid conflicts a DB table must be added with the following fields: 

Field | Description
-- | --
cleanup_time_interval | Minimum Interval to start the next cleanup routine
last_cleanup_start_timestamp | Start timestamp of the last cleanup routine
last_cleanup_end_timestamp | End timestamp of the last cleanup routine. If less than “last_cleanup_start_timestamp” it means the cleanup is still in progress

Each server will start the cleanup routine only if "last_cleanup_start_timestamp" + "cleanup_time_interval" <= current time. Once started, the process will follow the logic described in the diagram below:

Image

Note that to mark a folder as to delete an empty file named TO_DELETE is created inside the folder. Its creation timestamp will be used to check if the retention period has expired.

DevMattG:
This PR introduces the cleanup background routine to scan the store and remove obsolete folders (not referred by a record in the DB).

The following env variables are added to configure it:
- MOSAICOD_CLEANUP_TIME_INTERVAL (secs) -> specifies how often the routine should run. Default is 1 day.
- MOSAICOD_CLEANUP_RETENTION_DURATION (secs) -> how long a folder is retained before being permanently deleted. Default is 1 day.

A new table is added to the DB containing the logs for every cleanup run, keeping stats as marked and removed folders, as well as the folders for which some error occurred while trying to mark or delete them.

Closes #343


mosaicod/crates/mosaicod-facade/src/sequence.rs
```diff
@@ -81,31 +81,44 @@ pub async fn try_create(
...
+    if let Some(mdata) = &metadata {
+        record = record.with_user_metadata(mdata.clone());
     }
 
     let record = db::sequence_create(&mut tx, &record).await?;
 
     tx.commit().await?;
 
+    // 2. If metadata are present, save them to Store too.
+    if let Some(mdata) = metadata {
```
LuigiTer, line 103: This delete pattern is fragile: when the store write fails, we pay for two DB round-trips (insert + delete) for an operation that ultimately did nothing.

mosaicod/crates/mosaicod-facade/src/topic.rs
```diff
@@ -290,17 +288,25 @@ pub async fn writer(
...
+    // 1. Save path_in_store on DB.
     let mut cx = context.db.connection();
     db::topic_update_path_in_store(&mut cx, handle.id, path_in_store.clone()).await?;
 
+    // 2. Save metadata to Store.
+    let res =
+        metadata_write_to_store(&context, path_in_store.path_metadata().as_path(), mdata).await;
+
+    // Rollback: remove path_in_store from the topic db entry.
+    if let Err(e) = res {
```
LuigiTer, line 305: Same here. insert + delete in case of error while writing on store.

mosaicod/crates/mosaicod-task/src/cleanup.rs
```diff
@@ -0,0 +1,789 @@
...
+                path_in_store
+            );
+        }
+
+        Ok(false)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
LuigiTer, line 268: A scenario with 2 or more instance of the same server trying to cleanup on the same database, is possible?

mosaicod/crates/mosaicod-facade/src/sequence.rs
DevMattG: I think this depends on how often we expect writes to fail compared to how many 'create' requests with the same locator we receive. If we swap the store write with the DB record creation, every 'create' request will create a folder in the store before we check if the sequence already exists. This could lead to a malicious attack aimed at flooding the object storage.

mosaicod/crates/mosaicod-facade/src/topic.rs
DevMattG: Here we could think of revert the change and save the metadata file in the store as first step. There is already a check on the topic status at the beginning of the function that prevents other do_put  from flooding the object storage

mosaicod/crates/mosaicod-task/src/cleanup.rs
DevMattG: Ok, I'll try to add some more tests with 1+ server instances inside _tests_ folder.

mosaicod/crates/mosaicod-facade/src/sequence.rs
DevMattG: This is done even to prevent the newly created folder on the store from being marked as TO_DELETE just before the DB record creation.

mosaicod/crates/mosaicod-facade/src/topic.rs
DevMattG: Same here: this is done to prevent the newly created folder on the store from being marked as TO_DELETE just before the DB record update.

cabba (approved):

PR #496, opened 2026-05-13
Release/mosaico py/v0.4.3

pucciland95:
This release solves the problem of correctly reconstructing Mosaico `Serializable` objects when they have `MosaicoList` of `Serializable` objects as fields (i.e. FrameTransform has a MosaicoList of Tranform). 
In particular, before the fix, the reconstructed `Seriablizable` object had a `List[dict]` rather than a `List[Serializable]`. The bug was not picked up by the test since it used element-wise checking through `model_dump()` from `Pydantic` rather than type checking + element-wise checking.  

Therefore the changes are:

- Added type hinting to `list_` function: `return Annotated[list[source_type], arrow_list_type]`
- Modified [test](https://github.com/mosaico-labs/mosaico/blob/main/mosaico-sdk-py/src/testing/unit/test_message_arrow_roundtrip.py) removing `model_dump()` to assert object equality


fdicorato:
Please, move to draft until the SDK release is published on PyPi. Then, we will solve the conflicts by merging with main first and then the PR can be made.

fdicorato:
@pucciland95 the release has been tagged and published on PyPi.
Please, proceed with:
* merging with main (main -> release/mosaico-py/v0.4.3)
* conflicts resolution
* PR (release/mosaico-py/v0.4.3 -> main)

Thank you

fdicorato (approved):

PR #515, opened 2026-05-19
fix(mosaicod): Cleanup unit tests being stuck 

### LINKED ISSUE #510: bug(mosaicod): Cleanup unit tests being stuck (sometimes)
Message showed: `test_cleanup_1 has been running for over 60 seconds`.
**This happens sometimes (not systematic)**

LuigiTer:
The old shutdown signal worked like shouting "stop": only tasks that were already listening at that moment heard it; once shouted, the signal was gone for good.

Background tasks (cleanup, flight server) needed a brief moment after startup before they started listening, the cleanup task, for example, ran a full try_cleanup() first. If shutdown() was called during that window, the signal was shouted into an empty room and lost forever, and the task would keep sleeping and working as if nothing happened.

In serial test runs that window was tiny and never caught the signal too early. In parallel runs, filesystem and DB contention stretched the first try_cleanup() to seconds, opening the window wide enough to lose the signal and hang tests past the 60s threshold.

CancellationToken works like a flag that stays flipped: once cancel() is called, any task checking later, even much later, sees it immediately. The race is gone by construction.

- ShutdownNotifier public API (.shutdown(), .wait_for_shutdown(), Default, Clone) unchanged.
- inner() removed; .token() exposes a CancellationToken for use across crate boundaries (mosaicod-task cannot depend on mosaicod-server).
- Cleanup::run now takes CancellationToken directly.

Close #510

### DIFF
--- a/mosaicod/Cargo.toml
+++ b/mosaicod/Cargo.toml
@@ -59,7 +59,7 @@ sqlx = { version = "0.8.6", features = [
 ] }
 thiserror = "2.0.18"
 tokio = { version = "1.50.0", features = ["rt-multi-thread"] }
-tokio-util = { version = "0.7", features = ["rt"] }
+tokio-util = { version = "0.7" }
 url = "2.5.8"
 uuid = { version = "1.23.1", features = [ "v4" ] }
 ulid = "1.2.1"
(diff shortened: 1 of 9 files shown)


mosaicod/Cargo.toml
```diff
@@ -59,6 +59,7 @@ sqlx = { version = "0.8.6", features = [
 ] }
 thiserror = "2.0.18"
 tokio = { version = "1.50.0", features = ["rt-multi-thread"] }
+tokio-util = { version = "0.7", features = ["rt"] }
```
cabba: why we need `rt` feature?

LuigiTer: `rt` enables `sync` module which containts `CancellationToken`

cabba (approved):

Issue #481, opened 2026-05-05
feat(py): Implement Mosaico-to-ROS Sequence Egress (Reverse ROS Bridge)


fdicorato:
The current `ros_bridge` module is architected around the `RosbagInjector`, which translates ROS messages into the Mosaico Ontology for ingestion While this solves the archival problem, it does not support the "Targeted Debug" use case: extracting a specific time-windowed segment of a multi-topic, multi GB sequence into a `.mcap` or `.db3` file for playback in native ROS tools (e.g., Foxglove or RViz).

We need to implement a reverse pipeline that utilizes Mosaico’s **Streamers** and update the ROS Adapters to allow the reconstruction of ROS-compliant binary logs.

#### 1. Inverse Adaptation Layer
Currently, `ROSAdapterBase` subclasses implement `from_dict` to move data into Mosaico. We need to extend the adapter contract to include a reverse mapping:
* **New Method**: `to_ros(mosaico_msg: Message, ...) -> ROSType`
* **Purpose**: Convert strongly-typed Mosaico models (e.g., `IMU`, `Pose`) back into a ROS `msg_type` format expected by the `rosbags` serialization engine.

#### 2. The `SequenceExtractor` (Egress Orchestrator)
A new controller, similar to `RosbagInjector` should be implemented to manage the export lifecycle:
* **Stream Consumption**: Utilize `SequenceDataStreamer` to pull synchronized, chronologically ordered messages from the server.
* **Temporal Slicing**: Leverage the streamer’s `start_timestamp_ns` and `end_timestamp_ns` parameters to perform server-side pruning.
* **Topic Filtering** Leverage the streamer’s `topics` parameter to perform topic pruning.
* **Schema Reconstruction**: Map Mosaico `ontology_tag`s back to ROS `msg_type` strings using the existing `ROSBridge` registry.

#### 3. Integration with `rosbags` Writer
The extracted and translated data should be fed into the `rosbags`writer to support multiple output formats (`.mcap`, `.db3`).

### Target Workflow Example
```python
from mosaicolabs.ros_bridge import ROSSequenceExporter

# Goal: Export a 30-second window of a crash event to an MCAP file
with MosaicoClient.connect("localhost", 6726) as client:
    handler = client.sequence_handler("fleet_run_2026")
    
    exporter = ROSSequenceExporter(
        handler=handler,
        output_path="debug_segment.mcap",
        start_ns=1738508778000000000,
        end_ns=1738508808000000000,
[...]

PR #518, opened 2026-05-21
feat(py): implemented translation Mosaico->Rosbag

pucciland95:
# Summary

This PR aims at solving #481 and partially the discussion #61. 
In particular, the changes in this PR introduce a `SequenceExtractor` object that does the opposite operation of ROSInjestor: turns a Mosaico Sequence into a rosbag. 

## Changes

- for each `Adapter` a `to_ros()` function has been added translating a `Serializable`/`Message` into a `ros_msg` ready to be encoded into a `rosbag`.
- for each `Adapter` it is necessary to indicate whether it is the `default` adapter when returning from `MosaicoType` to ROS. One and one only adapter can be associated with a `MosaicoType`. To be the defaul Mosaico->ROS Adapter can be indicated as the following:

``` python3
@register_default_adapter(is_default=True)
class PoseAdapter(ROSAdapterBase[Pose]):
```

- introduced `MosaicoLoader` (mimicking `ROSLoader`) to load Mosaico Sequence
- introduced a `Loader` Protocol which both  `MosaicoLoader` and `ROSLoader` implement
- refactored ProgressManager. It now takes a `Loader` Protocol.
- Introduced a `SequenceExtractor` which uses a `MosaicoLoader` to turn Mosaico sequences into rosbags

## How to run

To run Sequence extractor example: 

1. Execute the backed using the script within `/Scripts`:

``` bash
./dev_env
```

2. Ingest the NVIDIA rosbag using the provided example:

``` bash
mosaicolabs.examples ros_injection --port 6276
```

3. Run the `SequenceExtractor` main file

``` bash
python3 sequence_extractor.py r2b_robotarm_0 --port 6276 --ros_distro ROS2_JAZZY --overwrite
```

The results should be a rosbag file that is reproducible using `ros2 bag play ...`

## Tests

**Several** new tests have been added to ensure that translating a data from Mosaico to ROS1/2 and viceversa works properly. Indeed a new unit test section can be found under `mosaico-sdk-py/src/testing/unit/ros_bridge/adapters`.

## Note

Currently targeted and tested `ROS_DISTRO` are:
- ROS1_NOETIC
- ROS2_DASHING
- ROS2_ELOQUENT
- ROS2_FOXY
- ROS2_GALACTIC
- ROS2_HUMBLE
- ROS2_IRON
- ROS2_JAZZY
- ROS2_KILTED


DevMattG:
@pucciland95 remember to fix the PR title!

mosaico-sdk-py/src/mosaicolabs/ros_bridge/adapters/geometry_msgs.py
```diff
@@ -150,7 +156,13 @@ def from_dict(cls, ros_data: dict) -> Pose:
             out_pose = cls.from_dict(pose_dict)
 
             # While unwinding recursion, attach metadata found at this level
-            out_pose.covariance = ros_data.get("covariance")
+            if ros_data.get("covariance"):
+                covariance = None
+                if _is_valid_covariance(ros_data.get("covariance")):
```
fdicorato: It could be better to call `get` only once, to avoid searching in a dictionary multiple times the same key:
```python
ros_covariance = ros_data.get("covariance")
if ros_covariance:
     ...
     if _is_valid_covariance(ros_covariance)
```

Please, do the same in all other places

mosaico-sdk-py/src/mosaicolabs/ros_bridge/adapters/geometry_msgs.py
```diff
@@ -878,12 +1306,74 @@ def from_dict(cls, ros_data: dict) -> Transform:
...
+
+        # Checking presence in typestore of requested message
+        if typestore.types.get(resolved_rosmsg_type) is None:
+            return None
+
+        # Unpacking Mosaico message / type
+        transform_data, ms_header = cls.unpack_mosaico_msg(mosaico_data)
+
+        # Filling the data
+        # source_frame_id = transform_data.source_frame_id or ""
```
fdicorato: `transform_data.source_frame_id` should not be put into `Header.frame_id` of such a transform?

mosaico-sdk-py/src/mosaicolabs/ros_bridge/loader.py
```diff
@@ -437,3 +470,209 @@ def __enter__(self):
...
+        self.requested_topics = topics
+        self.start_timestamp_ns = start_timestamp_ns
+        self.end_timestamp_ns = end_timestamp_ns
+
+        self.seq_handler = None
+        self.streamer = None
+
+    def validate_sequence(self):
+        if self.seq_handler is None:
+            all_seq = self.client.list_sequences()
```
fdicorato: No, avoid doing this. The number of sequences on the server can be huge! Just raise telling that the sequence handler cannot be retrieved, as the sequence may not exist

mosaico-sdk-py/src/mosaicolabs/ros_bridge/sequence_extractor.py
```diff
@@ -0,0 +1,556 @@
...
+    """
+
+    ros_distro: Optional[Stores] = None
+    """
+    The specific ROS distribution to use for message parsing (e.g., Stores.ROS2_HUMBLE). If None, defaults to Empty/Auto.
+
+    See [`rosbags.typesys.Stores`](https://ternaris.gitlab.io/rosbags/topics/typesys.html#type-stores).
+    """
+
+    storage_plugin: StoragePlugin = StoragePlugin.SQLITE3
```
fdicorato: Why defaulting to SQLITE3? I would set MCAP

mosaico-sdk-py/src/mosaicolabs/ros_bridge/sequence_extractor.py
```diff
@@ -0,0 +1,556 @@
...
+    mosaico_api_key: Optional[str] = None
+    """
+    The API key for authentication on the mosaico server. Defaults to None.
+    
+    If provided it must be have at least [`APIKeyPermissionEnum.Write`][mosaicolabs.enum.APIKeyPermissionEnum.Write]
+    permission.
+    """
+
+    tls_cert_path: Optional[str] = None
+    """Path to the TLS certificate file for secure connection on the mosaico server. Defaults to None."""
```
fdicorato: Add this to the docstring:
`If `tls_cert_path=None` and `enable_tls=True`, a standard one-way TLS (server authenticated only) connection is established`

mosaico-sdk-py/src/mosaicolabs/ros_bridge/sequence_extractor.py
```diff
@@ -0,0 +1,556 @@
...
+
+    enable_tls: bool = False
+    """Enable the TLS commmunication protocol. Defaults to False"""
```
fdicorato: improve docstring:
`Enable the TLS standard one-way TLS (server authenticated only) communication protocol. Defaults to False. If `tls_cert_path` is provided (not None), this flag does not have any effect.`

mosaico-sdk-py/src/testing/integration/helpers.py
```diff
@@ -89,7 +89,7 @@ def make_gps_msg(meas_time: Time):
             status=GPSStatus(
                 status=0,
                 service=2,
-                satellites=int(random.uniform(4, 20)),
+                # satellites=int(random.uniform(4, 20)),
```
fdicorato, line 92: why commented?

mosaico-sdk-py/src/testing/unit/ros_bridge/adapters/test_geometry_msgs.py
```diff
@@ -0,0 +1,1483 @@
...
+    return ROSMessage(
+        bag_timestamp_ns=100,
+        topic="/vector3",
+        msg_type="geometry_msgs/msg/Vector3Stamped",
+        data={"header": ros_header, "vector": vector3d.model_dump(exclude_none=True)},
+    )
+
+
+class TestVectoradapter:
+    def assert_vector3(self, vector3d: Vector3d, ros_msg: dict):
```
fdicorato, line 98: why not using `assert_vector3` in helpers.py?

fdicorato (changes requested):
Excellent job! Please respond to the comments and solve the merge conflicts

mosaico-sdk-py/src/testing/integration/helpers.py
pucciland95: Setting the `satellites` fields makes integration test `test_run_data_correctness` fail. The tests consists in: 

1) load a Sequence on the backend through the `inject_synthetic_sequence` (data for the Sequence are created using `make_gps_msg`).
2) take the Sequence and turn this into a rosbag
3) take the rosbag and use `translate()` on each data to reconstruct back the Mosaico Type and compare them with the original ones. 

Since the corresponding [ROS2 message](https://docs.ros2.org/foxy/api/sensor_msgs/msg/NavSatStatus.html) for `GPSStatus` does not have the `satellites` field, when passing from Mosaico to ROS type `satellites` field is dropped and never reconstructed back when returning to Mosaico Type. Consequently, the final assertion would fail since the original `satellites` field is not `None` .

Therefore, the only way to avoid the test failure is to comment out `satellites` field

pucciland95:
All comments have been addressed. Let me know if you think this PR still needs changes

fdicorato (approved):

PR #524, opened 2026-05-25
feat(mosaicod): Implement topic_filter_clusterize endpoint 

LuigiTer:
Implemented the entire path from retreiving topic handler to get the data stream, clusterize and stream back the result.

Fetches Parquet via predicate pushdown, streams filtered batches into clustering (optionally bounded by timestamp_range), forwards clusters as JSONL  with no buffering.

Returns empty stream on non-overlapping ranges or incoherent data. Special-cases clustering_dt_ns=0 as a single [min, max] cluster.

### DIFF
--- a/mosaicod/crates/mosaicod-ext/src/arrow_filter.rs
+++ b/mosaicod/crates/mosaicod-ext/src/arrow_filter.rs
@@ -1,5 +1,5 @@
 use arrow::array::AsArray;
-use arrow::datatypes::UInt64Type;
+use arrow::datatypes::Int64Type;
 use arrow::error::ArrowError;
 use arrow::record_batch::RecordBatch;
 use futures::{Stream, StreamExt};
@@ -133,13 +133,32 @@ fn extract_timestamps<'a>(
     }
 
     // unwrap here is safe because the timestamp's datatype is enforced in do_put session.
-    let timestamp_array = timestamp_array.as_primitive_opt::<UInt64Type>().unwrap();
-    Ok(timestamp_array.values())
+    let timestamp_array = timestamp_array
+        .as_primitive_opt::<Int64Type>()
+        .unwrap()
+        .values();
+
+    // SAFETY: reinterpreting &[i64] as &[u64] is sound here because:
+    //   - i64 and u64 have the same size (8 bytes) and alignment, so the
+    //     pointer cast preserves the memory layout of the slice;
+    //   - the timestamp column is enforced to be Int64 at ingest time
+    //     (do_put session), so we never reach this code with a different type;
+    //   - timestamps represent nanoseconds since the Unix epoch and are always
+    //     non-negative, hence every value fits in the positive range of i64
+    //     and maps one-to-one to u64 without any change in numeric meaning;
+    let ts: &[u64] = unsafe {
+        std::slice::from_raw_parts(
+            timestamp_array.as_ptr() as *const u64,
+            timestamp_array.len(),
+        )
+    };
+    Ok(ts)
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
+    use arrow::array::Int64Array;
     use arrow::array::UInt64Array;
     use arrow::datatypes::DataType;
     use arrow::datatypes::{Field, Schema};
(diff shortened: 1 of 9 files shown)


mosaicod/crates/mosaicod-ext/src/arrow_filter.rs
```diff
@@ -24,13 +24,44 @@ pub enum ClusteringError {
     Arrow(#[from] ArrowError),
 }
 
+pub fn clustering_error_to_status(e: ClusteringError) -> tonic::Status {
```
DevMattG: I'd put this inside `impl ClusteringError` and call it `to_status`

mosaicod/crates/mosaicod-ext/src/arrow_filter.rs
```diff
@@ -133,28 +171,51 @@ fn extract_timestamps<'a>(
     }
 
     // unwrap here is safe because the timestamp's datatype is enforced in do_put session.
-    let timestamp_array = timestamp_array.as_primitive_opt::<UInt64Type>().unwrap();
-    Ok(timestamp_array.values())
+    let timestamp_array = timestamp_array
```
DevMattG, line 176: Are we sure that negative timestamps are not allowed? @cabba

mosaicod/crates/mosaicod-ext/src/arrow_filter.rs
```diff
@@ -554,4 +633,57 @@ mod tests {
             assert_eq!(c.start_ns, (i as u64) * 1_000);
         }
     }
+
+    #[tokio::test]
+    async fn error_arrives_in_order_after_successful_clusters() {
+        // Primo batch: produce un cluster che si chiude (gap > dt)
```
DevMattG: Convert comments to english

mosaicod/crates/mosaicod-server/src/endpoint/actions/topic.rs
```diff
@@ -128,62 +131,127 @@ pub async fn notification_purge(ctx: &facade::Context, locator: String) -> Resul
...
+    ts: Option<types::TimestampRange>,
+    ontology_filter: query::OntologyFilter,
+) -> Result<SendableRecordBatchStream> {
+    let meta = facade::topic::metadata(context, handle).await?;
+    let format = meta.ontology_metadata.properties.serialization_format;
+    let batch_size = facade::topic::compute_optimal_batch_size(context, handle)
+        .await
+        .ok();
+
+    let path_in_store = handle
```
DevMattG, line 151: Use `Error::MissingDbData` instead of `Error::NotFound`

DevMattG (changes requested):

mosaicod/crates/mosaicod-ext/src/arrow_filter.rs
cabba: @pucciland95 do you have any experience with bags with negative timestamps?

pucciland95: According to [ROS timestamp documentation](https://docs.ros2.org/foxy/api/builtin_interfaces/msg/Time.html), timestamps are defined as:

```
int32 sec
uint32 nanosec
```
Therefore It could be possible to have negative numbers. However since it is defined as Unix Epoch time it should be always `>=0`. As far as my experience goes, I have never seen timestamps with negative values.

DevMattG (approved):

mosaicod/crates/mosaicod-server/src/endpoint/actions/topic.rs
```diff
@@ -128,62 +131,127 @@ pub async fn notification_purge(ctx: &facade::Context, locator: String) -> Resul
...
+    clustering_dt_ns: u64,
     ontology: Ontology,
     timestamp_range: Option<FilterTimestampRange>,
 ) -> Result<DoActionStream> {
     info!("filter clusterize for {}", locator);
 
-    if let Some(ts) = &timestamp_range {
-        ts.validate()?;
-    }
+    // 1. Validation and conversion to TimestampRange
```
cabba, line 178: Please remove numbered bullets like `1.`, `2.`

mosaicod/crates/mosaicod-server/src/endpoint/actions/topic.rs
```diff
@@ -128,62 +131,127 @@ pub async fn notification_purge(ctx: &facade::Context, locator: String) -> Resul
...
+    let topic_handle = facade::topic::Handle::try_from_locator(ctx, topic_locator).await?;
+    let timestamp_column = core::params::ARROW_SCHEMA_COLUMN_NAME_INDEX_TIMESTAMP.to_owned();
+    let ontology_filter = ontology.try_into()?;
+
+    // 5. RecordBatch stram filtered by timestamp if any and ontology
+    let batch_stream = query_by_timestamp(ctx, &topic_handle, ts, ontology_filter)
+        .await?
+        .map(|item| item.map_err(|e| ArrowError::ExternalError(Box::new(e))));
+
+    // 6. Channel setup
```
cabba, line 213: Describe a little better why we need a channel here

PR #527, opened 2026-05-26
ci: fixed github action workflow tag pattern matching

cabba:
Moved from `[0-9]+` to `*` in tag pattern matching for github workflows.


fdicorato (approved):
Great!

PR #530, opened 2026-05-28
feat(mosaicod): Implement missing operators when querying user_metadata fields

### LINKED ISSUE #511: Implement missing operators when querying `user_metadata` fields
We need to add support for the in and match operators for `user_metadata` fields within the query engine.

**Please note that these operators should only work for scalar fields (non-list, non-dictionary).**

### Examples
- `in` operator: `imu.acceleration.x in([1,6,-1]`) should evaluate to true if the value of `imu.acceleration.x` is either 1, 6, or -1.
- `match` operator: implement regex/pattern matching for scalar string values.

LuigiTer:
* Added the support to in and match operator for user metadata
* Added tests for query
* Added mosaico_regex_func in migrations for future changes in case of different regex extensions

Close #511


mosaicod/crates/mosaicod-db/migrations/20260528103216_functions.sql
```diff
@@ -0,0 +1,7 @@
+CREATE OR REPLACE FUNCTION mosaico_regex_func(value TEXT, pattern TEXT)
```
DevMattG: I would call it `mosaico_regex_match`

mosaicod/crates/mosaicod-query/src/filter.rs
```diff
@@ -129,7 +137,7 @@ impl IsSupportedOp for Value {
     }
 
     fn support_in(&self) -> bool {
-        matches!(self, Self::Boolean(_))
```
DevMattG, line 132: why not booleans?

mosaicod/tests/tests/query.rs
```diff
@@ -0,0 +1,501 @@
...
+    .await
+    .unwrap_err();
+
+    assert_eq!(err.code(), tonic::Code::Internal);
+
+    server.shutdown().await;
+}
+
+#[sqlx::test(migrator = "mosaicod_db::testing::MIGRATOR")]
+async fn test_query_in_with_booleans_is_rejected(pool: sqlx::Pool<db::DatabaseType>) {
```
DevMattG: I would allow this as a valid request. This is not the best way to check a boolean for sure, but I don't see any drawback in leaving it possible. It will be up to the user to decide

mosaicod/tests/tests/query.rs
```diff
@@ -0,0 +1,501 @@
...
+            "topic": { "user_metadata": { "x": { "$in": [] } } }
+        }),
+    )
+    .await;
+
+    assert!(
+        result.is_err(),
+        "empty $in should not silently return results"
+    );
+
```
DevMattG, line 308: Add a check for the error code as in the previous test

mosaicod/tests/tests/query.rs
```diff
@@ -0,0 +1,501 @@
...
+    let res = actions::query(
+        &mut client,
+        json!({
+            "topic": { "user_metadata": { "vehicle": { "$match": "" } } }
+        }),
+    )
+    .await
+    .unwrap_err();
+
+    assert_eq!(res.code(), tonic::Code::Internal);
```
DevMattG, line 515: This is not an internal error. The problem is related to an erroneous input provided by the user. I'd use the `InvalidArgument` tonic code

DevMattG (changes requested):

mosaicod/crates/mosaicod-query/src/filter.rs
LuigiTer: Do we want to accept query like `bool_value in [true, false]` ?

cabba (approved):

DevMattG (approved):

mosaicod/tests/tests/query.rs
LuigiTer: This will be solved in issue #532

PR #539, opened 2026-05-29
fix(mosaicod): Field ontology tag is not validated

### LINKED ISSUE #535: Field ontology tag is not validated
When building a query (clusterize) that filters on a field path, the prefix segment (i.e., the part before the first dot, interpreted as the ontology tag) is never validated against any known set of allowed values.

This means that querying `imu.acceleration.x` and `foo.bar.x` produce the same result, even though `foo` is not a valid ontology tag and `bar` is not a valid schema path. The prefix is silently accepted and discarded, while only the trailing segment(s) after the first dot are used.

LuigiTer:
The DataFusion filter in query_by_timestamp was silently ignoring the ontology tag prefix, allowing any arbitrary prefix to match valid fields.

Added validation that rejects filter tags not matching the topic's registered ontology tag, and fixed the test helper to use the correct tag.

Close #535

### DIFF
--- a/mosaicod/crates/mosaicod-query/src/filter.rs
+++ b/mosaicod/crates/mosaicod-query/src/filter.rs
@@ -388,6 +388,10 @@ impl OntologyFilter {
         self.ontology.get(field)
     }
 
+    pub fn ontology_tags(&self) -> impl Iterator<Item = &str> + '_ {
+        self.ontology.keys().map(|f| f.ontology_tag())
+    }
+
     /// Exports filter data as a unique expression group
     pub fn into_expr_group(self) -> OntologyExprGroup<Value> {
         OntologyExprGroup {
--- a/mosaicod/crates/mosaicod-query/src/filter.rs
+++ b/mosaicod/crates/mosaicod-query/src/filter.rs
@@ -388,6 +388,7 @@ impl OntologyFilter {
         self.ontology.get(field)
     }
 
+    /// Returns an iterator over the ontology tags.
     pub fn ontology_tags(&self) -> impl Iterator<Item = &str> + '_ {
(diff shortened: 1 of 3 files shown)


mosaicod/crates/mosaicod-query/src/filter.rs
```diff
@@ -388,6 +388,10 @@ impl OntologyFilter {
         self.ontology.get(field)
     }
 
+    pub fn ontology_tags(&self) -> impl Iterator<Item = &str> + '_ {
```
cabba, line 392: Add doc

mosaicod/crates/mosaicod-server/src/endpoint/actions/topic.rs
```diff
@@ -144,6 +144,17 @@ pub async fn query_by_timestamp(
 ) -> Result<SendableRecordBatchStream> {
     let meta = facade::topic::metadata(context, handle).await?;
     let format = meta.ontology_metadata.properties.serialization_format;
+    let topic_tag = &meta.ontology_metadata.properties.ontology_tag;
+
+    // check if filter tag is registered before execute query
+    for filter_tag in ontology_filter.ontology_tags() {
+        if filter_tag != topic_tag {
+            return Err(core::Error::bad_request(format!(
+                "unknown ontology tag {filter_tag}, topic uses {topic_tag}"
```
cabba:
```suggestion
                "wrong ontology tag {filter_tag}, topic uses {topic_tag}"
```

cabba: Since we can have the onology tag resigered, but for another topic.

cabba (changes requested):

cabba (approved):

PR #543, opened 2026-06-01
feat(mosaicod): Add regex support for sequence and topic name using match operator 

LuigiTer:
Replace LIKE-based matching with mosaico_regex_match for topic and sequence locator name filters, aligning their behaviour with the existing user_metadata $match operator.

### DIFF
--- a/mosaico-sdk-py/src/testing/integration/test_query_regex.py
+++ b/mosaico-sdk-py/src/testing/integration/test_query_regex.py
@@ -1,0 +1,154 @@
+import pytest
+from pyarrow import ArrowInvalid
+from pyarrow.flight import FlightError
+
+from mosaicolabs.comm import MosaicoClient
+from mosaicolabs.models.query import QuerySequence, QueryTopic
+from testing.integration.config import (
+    UPLOADED_GPS_TOPIC,
+    UPLOADED_IMU_CAMERA_TOPIC,
+    UPLOADED_IMU_FRONT_TOPIC,
+    UPLOADED_SEQUENCE_NAME,
+)
+
+
+def test_query_topic_name_regex_start_anchor(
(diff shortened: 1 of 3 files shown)


cabba:
@fdicorato please review the python tests.

@LuigiTer added some full-stack tests to lock in this behavior and avoid breaking changes in the future.

mosaico-sdk-py/src/testing/integration/test_query_regex.py
```diff
@@ -0,0 +1,154 @@
...
+    with pytest.raises(FlightError):
+        mosaico_client.query(QuerySequence().with_name_match("((unclosed"))
+
+    mosaico_client.close()
+
+
+def test_query_sequence_with_percentage(
+    mosaico_client: MosaicoClient, inject_synthetic_sequence
+):
+    query_resp = mosaico_client.query(QuerySequence().with_name_match("%"))
```
fdicorato, line 142: what are we testing here? `%` is not a valid operator?

fdicorato (approved):
Approved for Python part. Just a question about the 'percent' test case.

mosaico-sdk-py/src/testing/integration/test_query_regex.py
LuigiTer: In Posix regular expressions, the `%` symbol is considered a literal character. This test verifies that it is accepted as a character, but an empty result is expected because we don't accept locators with `%` in their names.

cabba (approved):

mosaico-sdk-py/src/testing/integration/test_query_regex.py
fdicorato: Great! Thank you

PR #547, opened 2026-06-03
feat(mosaicod): enabled * and ** glob patterns for user metadata queries

### LINKED ISSUE #528: Enable * and **  glob wildcards when addressing user metadata keys
The `*` wildcard is the operator for a single-level wildcard in JSON path queries. We want to consent its usage on user metadata.

Examples:

Pattern | Matches
-- | --
*.status | step1.status, step2.status
\*.\*.status | action.pick.status, action.move.status, step1.substep.status
action.*.status | action.pick.status, action.move.status
action[*].status | action[0].status, action[1].status

DevMattG:
To enable it we switched to Json path query language.

__Note__: `match` operator does not support it yet (see #549).

Closes #528


mosaicod/crates/mosaicod-db/src/sql/pg_queries/compilers.rs
```diff
@@ -205,111 +191,136 @@ mod internal {
...
                     if let query::Value::Text(text) = value {
                         if text.is_empty() {
                             return Err(query::Error::empty_pattern(field.to_owned()));
                         }
+
+                        let placeholder = self.consume_placeholder();
+
+                        let subfield = format!(
+                            "{{{}}}",
+                            field.chars().skip(2).collect::<String>().replace(".", ",")
```
LuigiTer: If I understand correctly, this `skip(2)` is used to skip `$.` but I wouldn't do it at the query level but rather with `flied.strip_prefix("$.")` to avoid silent errors in case `fmt_clause` will change from `$.` to something like `$[`

mosaicod/crates/mosaicod-db/src/sql/pg_queries/compilers.rs
```diff
@@ -205,111 +191,136 @@ mod internal {
...
                 query::Op::Eq(v) => {
                     let v: query::Value = v.into();
+                    let placeholder = self.consume_placeholder();
                     query::CompiledClause::new(
                         format!(
-                            "{} = {}",
-                            self.fmt_value(field, &v),
-                            self.consume_placeholder()
+                            "jsonb_path_exists({}, '{} ? (@ == $val)', jsonb_build_object('val', {}))",
+                            self.field, field, placeholder
```
LuigiTer, line 200: `field` here is passed directly inside the query, if the input will be something like `imu.acceleration.x' OR TRUE --` what happens? It is a possibile sql injection.

LuigiTer (changes requested):

mosaicod/crates/mosaicod-db/src/sql/pg_queries/compilers.rs
DevMattG: I think we need to check this more deeply and in other parts too. I would create a separate issue.

LuigiTer (approved):

mosaicod/crates/mosaicod-db/src/sql/pg_queries/compilers.rs
cabba: This behaviour will be addressed in issue #559

cabba (approved):

PR #556, opened 2026-06-08
feat(cli): add cli to interact with mosaico daemon

mattiaferrari02:
This PR adds CLI project for mosaico python SDK


pucciland95:
Hi @mattiaferrari02 and thank you for contributing. 
The CI is failing the static analysis check. Probably the code has not been formatted using `Ruff`.

Please follow our [contribution guidelines](https://docs.mosaico.dev/python-sdk/code_contributing/) and activate `pre-commit` to avoid this. Once `pre-commit` has been activated run:

```
poetry run pre-commit run --all-files
```

to correctly format your code

Cheers,

Niccolo

mosaico-sdk-py/src/mosaicolabs_cli/utils/mosaico_profile.py
pucciland95, line 1: File name (MosaicoProfile.py) is CamelCase when should be snake_case (mosaico_profile.py). Is there a particular reason for this?

mosaico-sdk-py/src/mosaicolabs_cli/utils/MosaicoProfile.py
```diff
@@ -0,0 +1,157 @@
...
+        config_data = load_config(config_path)
+
+        profile_dict: Dict[str, Any] = {}
+
+        if profile_name:
+            content = config_data.get(profile_name)
+            if not isinstance(content, dict):
+                console.print(
+                    f"[bold red]Error:[/bold red] Profile '{profile_name}' not found in configuration."
+                )
```
pucciland95: Missmatch between console.print() output and what if statement is checking. Maybe what you meant is:

```suggestion
            if not isinstance(content, dict) or content is None:
                console.print(
                    f"[bold red]Error:[/bold red] Profile '{profile_name}' not found in configuration."
                )
```

mosaico-sdk-py/src/mosaicolabs_cli/utils/MosaicoProfile.py
```diff
@@ -0,0 +1,157 @@
...
+
+        # Resolve TLS: env > config
+        if env_tls is not None:
+            tls = env_tls.lower() in ("1", "true", "yes")
+        else:
+            tls = bool(profile_dict.get("tls", False))
+
+        return cls(
+            host=env_host or profile_dict.get("host", ""),
+            port=6276 if env_host else int(profile_dict.get("port", 6276)),
```
pucciland95: I think it would be better to do not hardcode this value but make it a constant variable (i.e. DEFAULT_MOSAICO_DEV_PORT) so we can change it with ease. What do you think?

pucciland95: Furthermore, I do not get why the port is always 6276 when `env_host` is defined. Shouldn't the port be extracted from the `env_host` first and if not found fallback to `6276`?

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
```diff
@@ -0,0 +1,305 @@
...
+            host = typer.prompt("Mosaico Server Host")
+        if port is None:
+            port_input = typer.prompt(
+                "Mosaico Server Port (leave empty for 6726)", default=""
+            )
+            port_input = (
+                port_input.strip() if isinstance(port_input, str) else port_input
+            )
+            if port_input == "":
+                port = 6726
```
pucciland95: Probably it would be better to have a contant defining it like DEFAULT_MOSAICO_PORT

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
```diff
@@ -0,0 +1,305 @@
...
+        for profile_name, profile_content in config_data.items():
+            if profile_name != name and isinstance(profile_content, dict):
+                profile_content["default"] = False
+
+    if is_default or not config_data:
+        is_default = True
+        for profile in config_data.values():
+            if isinstance(profile, dict) and "default" in profile:
+                profile["default"] = False
+
```
pucciland95: I am confused by this double check that where, in both cases, sets all the ["default"] key to `False`. Could you please clarify why both are necessary?

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
```diff
@@ -0,0 +1,305 @@
...
+    config_data[name] = {
+        "host": host,
+        "port": port,
+        "api_key": api_key.strip() if api_key else "",
+        "tls": tls,
+        "cert_path": cert_path.strip() if cert_path else "",
+        "default": should_be_default,
+    }
+
```
pucciland95, line 127: In this way, if an existing profile with the same name (and possibly different parameters) exists, it would be overridden. Probably the user should be:

1) warned that an existing profile with the same name already exists and we should ask him/her if he/she is sure to override the existing one

2) informed that adding the profile would not be possible since it already exists

what do you think?

pucciland95 (changes requested):
Thank you for the contribution. I still need to look at other files, but in the meanwhile consider the comments I have left 😃

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
mattiaferrari02: my bad, we don't need both

mosaico-sdk-py/src/mosaicolabs_cli/utils/MosaicoProfile.py
mattiaferrari02: you are absolutely right

mosaico-sdk-py/src/mosaicolabs_cli/utils/mosaico_profile.py
mattiaferrari02: Oops, force of habit from TypeScript! Renaming to `mosaico_profile.py` right away

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
mattiaferrari02: Ok, with interactive mode let's prompt the user for the confirmation, otherwise in non-interactive mode let's just print a warning that the profile is being overwritten.

Maybe we could also throw an error in non interactive mode and to overwrite the profile in that particular use case we can add a `--force` option to the command

pucciland95: I like this:

> Maybe we could also throw an error in non interactive mode and to overwrite the profile in that particular use case we can add a --force option to the command

DevMattG (approved):

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
mattiaferrari02: done!

mosaico-sdk-py/src/mosaicolabs_cli/utils/mosaico_profile.py
```diff
@@ -0,0 +1,165 @@
...
+        """Strip protocol prefix and extract embedded port if present."""
+        h = host.strip().rstrip("/")
+        h = h.replace("http://", "").replace("https://", "")
+
+        if port == DEFAULT_MOSAICO_PORT and ":" in h:
+            host_part, port_part = h.rsplit(":", 1)
+            if port_part.isdigit():
+                return host_part, int(port_part)
+
+        return h, port
```
pucciland95: I am not convinced by this function. For example: I don't like that to get the port from the host (e.g. https://localhost:1234 I should get 1234) I need to remember to set the `port` parameter to `DEFAULT_MOSAICO_PORT`. 

I propose something like this (feel free to further modify the code as you prefer):

```suggestion
    @staticmethod
    def _normalize_host_port(host: str, port: Optional[int] = None) -> tuple[str, int]:
        """
        Strip protocol prefix and extract embedded port if present.
        A port may come from the `port` argument or be embedded in `host`
        (e.g. "localhost:1234"), but not both. If neither is given, falls
        back to DEFAULT_MOSAICO_PORT.
        """
        # Input cleaning
        h = host.strip().rstrip("/")
        h = h.replace("http://", "").replace("https://", "")

        # Extract host and port (if present)
        if ":" in h:
            h, embedded_port = h.rsplit(":", 1)

            if not embedded_port.isdigit():
                raise ValueError("Malformed input host {h}")

        # Check double port definition
        if embedded_port is not None and port is not None: # (i.e. localhost:1234 port=4567)
            raise ValueError(
                f"Port given twice"
            )

        final_port = embedded_port if embedded_port is not None else port

        return h, final_port if final_port is not None else DEFAULT_MOSAICO_PORT
```

what do you think?

mosaico-sdk-py/src/mosaicolabs_cli/utils/mosaico_profile.py
```diff
@@ -0,0 +1,165 @@
...
+        host: str = "",
+        port: int = DEFAULT_MOSAICO_PORT,
+        api_key: str = "",
+        tls: bool = False,
+        cert_path: str = "",
+    ):
+        self.host, self.port = self._normalize_host_port(host, port)
+        self.api_key = api_key.strip() if api_key else ""
+        self.tls = tls
+        self.cert_path = cert_path.strip() if cert_path else ""
```
pucciland95, line 33: What about adding also the default flag using a `bool`?

pucciland95: And probably also its `name`

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
```diff
@@ -0,0 +1,320 @@
...
+                table.add_row(name, p.host, str(p.port), default_marker)
+
+        console.print(table)
+
+    elif output == OutputFormat.CSV:
+        for name, content in config_data.items():
+            if isinstance(content, dict):
+                p = MosaicoProfile.from_dict(content)
+                is_default = str(content.get("default", False)).lower()
+                console.print(f"{name},{p.host},{p.port},{is_default}")
```
pucciland95: Adding `_is_default` and name to `MosaicoProfile` would simplify this to (after updating `__repr()__`):


```suggestion
                console.print(f"{p}")
```

mosaico-sdk-py/src/mosaicolabs_cli/utils/mosaico_profile.py
```diff
@@ -0,0 +1,165 @@
...
+        # Resolve TLS: env > config
+        if env_tls is not None:
+            tls = env_tls.lower() in ("1", "true", "yes")
+        else:
+            tls = bool(profile_dict.get("tls", False))
+
+        if env_host:
+            resolved_host, resolved_port = cls._normalize_host_port(
+                env_host, DEFAULT_MOSAICO_PORT
+            )
```
pucciland95: I think this may confuse the reader and it is the reason why `_normalize_host_port()` needs refactoring. 
If I pass the `port` parameter I expect that the parameter is set as port. In this case instead, passing `DEFAULT_MOSAICO_PORT` is a key argument to say "look what's in `env_host` and ignore the port param I passed you". With the new code this will look like:


```suggestion
            resolved_host, resolved_port = cls._normalize_host_port(env_host)
```

and the cases are:

1) input: "https://localhost:1234" -> host=localhost, port=1234
2) input: "https://localhost" -> host=localhost, port=`DEFAULT_MOSAICO_PORT`
3) input: "https://localhost:random" -> raise ValueError()

mosaico-sdk-py/src/mosaicolabs_cli/utils/config.py
```diff
@@ -0,0 +1,76 @@
...
+    TABLE = "table"
+    CSV = "csv"
+
+
+def get_config_path() -> Path:
+    """Resolve the configuration file path based on the environment or default."""
+    env_path = os.getenv(MosaicoEnv.CONFIG_PATH)
+    if env_path:
+        return Path(env_path)
+    return Path.home() / ".mosaico" / "config.toml"
```
pucciland95: Set constants: 
1) `DEFAULT_CONFIG_FOLDER`=".mosaico" 
2) `DEFAULT_CONFIG_FILENAME`="config.toml"
2) `DEFAULT_CONFIG_PATH`= `DEFAULT_CONFIG_FOLDER` / `DEFAULT_CONFIG_FILENAME`

What do you think?

mosaico-sdk-py/src/mosaicolabs_cli/utils/config.py
```diff
@@ -0,0 +1,76 @@
...
+    if not path.exists():
+        return {}
+    try:
+        with open(path, "rb") as f:
+            return tomllib.load(f)
+    except Exception as e:
+        error_console.print(
+            f"[bold red]Error:[/bold red] Failed to parse config file: {e}"
+        )
+        raise typer.Exit(code=1)
```
pucciland95: I would modify this as the following to allow passing no path. What do you think?

```suggestion
def load_config(path: Optinal[Path]) -> dict:
    """
        Safely load existing TOML configuration or return an empty dict if missing.
        If no `path` is passed, the default one is used 
    """

    if not path:
        path = DEFAULT_CONFIG_PATH

    if not path.exists():
        return {}
    try:
        with open(path, "rb") as f:
            return tomllib.load(f)
    except Exception as e:
        error_console.print(
            f"[bold red]Error:[/bold red] Failed to parse config file: {e}"
        )
        raise typer.Exit(code=1)
```

mosaico-sdk-py/src/mosaicolabs_cli/commands/sequence.py
```diff
@@ -0,0 +1,219 @@
...
+
+            results = client.query(query)
+
+            if not results:
+                spinner.stop()
+                console.print("No sequences found matching the criteria.")
+                raise typer.Exit()
+
+            if limit:
+                results: QueryResponse = results[:limit]
```
pucciland95: Type annotation here should be `list[QueryResponseItem]`

pucciland95: Furthrmore, I would avoid shadowing original response that is the actual `QueryResponse`

mosaico-sdk-py/src/mosaicolabs_cli/commands/sequence.py
```diff
@@ -0,0 +1,219 @@
...
+
+            rows = []
+            for item in results:
+                handler = client.sequence_handler(item.sequence.name)
```
pucciland95, line 112: Please add:


```suggestion
                handler = client.sequence_handler(item.sequence.name)
                if handler is None:
                    continue
```

I know that this should not happen, but Pylance complains that `handler` may be `None` since `client.sequence_handler(item.sequence.name)` may return `None`

pucciland95 (changes requested):

pucciland95:
Please find some comments above. 
I think some changes need some discussion (like the one about `Profiles`) and I would like your opinion about that.
 
I would also consider creating a `singleton` with a registry that handles Profiles 'Reading'/'Writing' from the `.toml` file but I think we could keep that for another PR

mosaico-sdk-py/src/mosaicolabs_cli/utils/mosaico_profile.py
mattiaferrari02: Yeah makes more sense this way

mosaico-sdk-py/src/mosaicolabs_cli/utils/config.py
mattiaferrari02: Yes, much better

mosaico-sdk-py/src/mosaicolabs_cli/utils/config.py
mattiaferrari02: Ok we can modify it this way, but at this point we should also add support for another env variabile MOSAICO_CONFIG_PATH so that users can use a custom profile path.

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
mattiaferrari02: But the __repr__ for `MosaicoProfile` class doest not produce a csv-like string, if we want to simplify this output we could define a new function to output the class props as csv (something like `to_csv_row` maybe)

mosaico-sdk-py/src/mosaicolabs_cli/utils/mosaico_profile.py
mattiaferrari02: Ok i'll refactor this function to behave like this. It should be more clear this way

mosaico-sdk-py/src/mosaicolabs_cli/utils/mosaico_profile.py
mattiaferrari02: Makes sense

mattiaferrari02:
> Please find some comments above. I think some changes need some discussion (like the one about `Profiles`) and I would like your opinion about that.

I will apply your suggested changes about `MosaicoProfile` class:
- refactor `_normalize_host_port`
- add a function to print the class props as csv instead of using `__repr__`

And i will add compatibility with a new env var MOSAICO_CONFIG_PATH

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
pucciland95: Oh ok, I did not get that you wanted a CSV like string. Therefore, the `to_csv_row` you propose sounds perfect

mattiaferrari02: That's because that line of code will be called if a user prints his profiles in csv output format

mosaico-sdk-py/src/mosaicolabs_cli/utils/config.py
pucciland95: Doesn't the `MosaicoEnv` class already contains such env variable?

``` python
class MosaicoEnv(str, Enum):
    """Canonical environment variable names for the Mosaico platform."""

    DAEMON_URL = "MOSAICO_DAEMON_URL"
    API_KEY = "MOSAICO_API_KEY"
    TLS = "MOSAICO_TLS"
    CERT_PATH = "MOSAICO_CERT_PATH"
    PROFILE = "MOSAICO_PROFILE"
    CONFIG_PATH = "MOSAICO_CONFIG_PATH" # <-------- THIS
```

mattiaferrari02: I added it just now

mattiaferrari02:
Ok i should have fixed tests

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
```diff
@@ -0,0 +1,321 @@
...
+    """
+    Add or update a connection profile inside the configuration file.
+    """
+
+    config_path = get_config_path()
+    config_data = load_config(config_path)
+
+    if interactive:
+        if not host:
+            host = typer.prompt("Mosaico Server Host")
```
pucciland95: It would make sense to ensure the user input is consitent using:


```suggestion
            host = typer.prompt("Mosaico Server Host", type=str)
```

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
```diff
@@ -0,0 +1,321 @@
...
+        if port is None:
+            port_input = typer.prompt(
+                f"Mosaico Server Port (leave empty for {DEFAULT_MOSAICO_PORT})",
+                default="",
+            )
```
pucciland95: It would make sense to ensure the user input is consitent using:

```suggestion
            port_input = typer.prompt(
                f"Mosaico Server Port (leave empty for {DEFAULT_MOSAICO_PORT})",
                default="",
                type=int
            )
```

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
```diff
@@ -0,0 +1,321 @@
...
+            else:
+                try:
+                    port = int(port_input)
+                except Exception:
+                    error_console.print(
+                        f"[bold red]Error:[/bold red] Invalid port '{port_input}'. Expected an integer."
+                    )
+                    raise typer.Exit(code=1)
+        if not api_key:
+            api_key = typer.prompt("API Key", hide_input=True, default="")
```
pucciland95: It would make sense to ensure the user input is consitent using:

```suggestion
            api_key = typer.prompt("API Key", hide_input=True, default="", type=str)
```

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
```diff
@@ -0,0 +1,321 @@
...
+        if not tls:
+            tls = typer.confirm("Enable TLS?", default=False)
+        if tls and not cert_path:
+            cert_path = typer.prompt(
+                "TLS CA Certificate Path (optional, press Enter to skip)", default=""
+            )
```
pucciland95, line 80: It would make sense to ensure the user input is consitent using:


```suggestion
            cert_path = typer.prompt(
                "TLS CA Certificate Path (optional, press Enter to skip)", default="", type=str
            )
```

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
```diff
@@ -0,0 +1,321 @@
...
+                f"[bold red]Error:[/bold red] Profile [yellow]'{name}'[/yellow] already exists. "
+                f"Use `--force` to overwrite."
+            )
+            raise typer.Exit(code=1)
+
+    was_already_default = False
+    if name in config_data and isinstance(config_data[name], dict):
+        was_already_default = config_data[name].get("default", False)
+
+    should_be_default = is_default or not config_data or was_already_default
```
pucciland95: `config_data` has been accessed several times here. I think it should be safe to remove the check:


```suggestion
    should_be_default = is_default or or was_already_default
```

or am I missing something?

mosaico-sdk-py/src/mosaicolabs_cli/commands/sequence.py
```diff
@@ -0,0 +1,233 @@
...
+                    continue
+                metadata_str = ", ".join(_flatten_metadata(handler.user_metadata))
+                rows.append(
+                    (
+                        item.sequence.name,
+                        str(handler._timestamp_ns_min),
+                        str(handler._timestamp_ns_max),
+                        metadata_str,
+                    )
+                )
```
pucciland95, line 124: I would consider adding the `handler.created_timestamp` so to combine this `ls` command with another `ls` that can then use the `--created-after` or `--created-after` flags

mosaico-sdk-py/src/mosaicolabs_cli/utils/mosaico_profile.py
```diff
@@ -0,0 +1,193 @@
...
+    def resolve(cls, profile_name: Optional[str] = None) -> "MosaicoProfile":
+        """
+        Resolve a MosaicoProfile applying the full precedence chain.
+        """
+        env_host = os.getenv(MosaicoEnv.DAEMON_URL)
+        env_api_key = os.getenv(MosaicoEnv.API_KEY)
+        env_tls = os.getenv(MosaicoEnv.TLS)
+        env_cert_path = os.getenv(MosaicoEnv.CERT_PATH)
+
+        config_data = load_config()
```
pucciland95: Shouldn't this be:


```suggestion
            config_path = get_config_path()
            config_data = load_config(config_path)
```

mosaico-sdk-py/src/mosaicolabs_cli/utils/mosaico_profile.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def to_dict(self) -> Dict[str, Any]:
+        return {
+            "host": self.host,
+            "port": self.port,
+            "api_key": self.api_key,
+            "tls": self.tls,
+            "cert_path": self.cert_path,
+            "default": self.is_default,
+        }
```
pucciland95, line 122: Missing `name` attribute or this is the intended behaviour?

pucciland95 (changes requested):
Some more comments. We are almost there :D. Before closing I will add some other issues that will be taken into account for later PRs

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
mattiaferrari02: The check for the default works only if we check also for "not config_data"

mosaico-sdk-py/src/mosaicolabs_cli/commands/sequence.py
```diff
@@ -0,0 +1,235 @@
...
+        table.add_column("User Metadata", style="green", width=35)
+
+        for row in rows:
+            table.add_row(*row)
+
+        console.print(table)
+
+    else:
+        for name, created, ts_min, ts_max, _ in rows:
+            print(f"{name},{created},{ts_min},{ts_max}")
```
pucciland95: Notice that returning this (name,created,ts_min,ts_max) would break all the other commands receiving data through pipe `|`. I just tested this:

`mosaico sequence ls --locator "r2b_galileo_0" | mosaico sequence stat`

and returned the following error:

`Fetching sequences...Error: Invalid handler format 'r2b_galileo_0,1782381260690877255,1707938716174752655,1707938738174395767'. Expected 'name,tsmin,tsmax'`

mosaico-sdk-py/src/mosaicolabs_cli/commands/topic.py
```diff
@@ -0,0 +1,339 @@
...
+
+
+@app.command(name="ls")
+def list_topics(
+    ctx: typer.Context,
+    locator: Optional[str] = typer.Option(
+        None,
+        "--locator",
+        help="Locator search pattern (supports glob-style * wildcards).",
+    ),
```
pucciland95, line 41: According to the design document, mosaico topic ls should have a `--sequence` flag that is missing. This is useful to get topics from a specific Mosaico sequence. It should support wildcard as the `--locator` flag

mosaico-sdk-py/src/mosaicolabs_cli/commands/sequence.py
mattiaferrari02: Ok i see, this is a change that i added during this review flow, it is a quite big error, i'll fix

mosaico-sdk-py/src/mosaicolabs_cli/commands/topic.py
mattiaferrari02: Yeah but you can just use the --locator to filter by `<sequence-Name>/*`

pucciland95 (approved):

pucciland95:
@fdicorato @cabba

PR #565, opened 2026-06-11
refactor!: changed "locator" field into "name" inside queries

### LINKED ISSUE #529: Obsolete "name" field inside query doc
Inside `Query` section of the doc, the first example under `Syntax` paragraph is using the `name` key for the sequence instead of `locator`.

Besides, the `locator` key for the topic must be replaced by `name` both in the code and the doc (where used)

DevMattG:
Closes #529


fdicorato (approved):
Changes to Python code approved

cabba (approved):

PR #577, opened 2026-06-12
feat(py): ROS msg ENUM and type loaded into Mosaico topic metadata

### LINKED ISSUE #568: feat: missing  ENUM and original ros_msgtype within Mosaico topic metadata
## Problem 
As explained in #61 , when traslating from ROS->Mosaico we miss to bring the value of the `ROS ENUM` (if present within ROS message definition) and the original `ros_msgtype`.

1) `ROS ENUM` within metadata would ease the understanding of the timeseries value meaning. For example, `power_supply_status` may get values:

| Value | Meaning |
| -------- | ------- |
|   `0`    | POWER_SUPPLY_STATUS_UNKNOWN
|   `1`    | POWER_SUPPLY_STATUS_CHARGING
|   `2`    | POWER_SUPPLY_STATUS_DISCHARGING

these meaning should be carried over to Mosaico to later understand their meaning

2) `ros_msgtype` on the other hand, would ease the passage Mosaico->ROS since it allows the correct reconstruction of the ROS original type (now falling back always to default)

## Solution

1) This can be solve inspecting the `typestore` during the `ROSLoader` parsing and inject the metadata at Mosaico topic creation.

2) During `enum` metadata creation, the `ros_msgtype` information can be retrieved and integrated.

The expected final metadata is:

``` json

{
  "_ros_": 
  {
      "enum": {
            "POWER_SUPPLY_STATUS_UNKNOWN": 0,
            "POWER_SUPPLY_STATUS_CHARGING": 1,
            "POWER_SUPPLY_STATUS_DISCHARGING": 2
       },
      "ros_msgtype": "sensor_msgs/msg/BatteryState"
  }
}

```

pucciland95:
After this PR, the Mosaico topic metadata for each ROS topic will be built automatically from two parts:

1) `ROS Enums`: the fixed constant values defined by the ROS message type
2) origianal ROS message type: the full type name of the source message

as an example metadata for a [sensor_msgs/msg/BatteryState](https://docs.ros.org/en/noetic/api/sensor_msgs/html/msg/BatteryState.html) topic would be:

``` json
"_ros_":
{
    "enums":
    {
        "POWER_SUPPLY_STATUS_UNKNOWN": 0,
        "POWER_SUPPLY_STATUS_CHARGING": 1,
        "POWER_SUPPLY_STATUS_DISCHARGING": 2,
        "POWER_SUPPLY_STATUS_NOT_CHARGING": 3,
        "POWER_SUPPLY_STATUS_FULL": 4,
        "POWER_SUPPLY_HEALTH_UNKNOWN": 0,
        "POWER_SUPPLY_HEALTH_GOOD": 1,
        "POWER_SUPPLY_HEALTH_OVERHEAT": 2,
        "POWER_SUPPLY_HEALTH_DEAD": 3,
        "POWER_SUPPLY_HEALTH_OVERVOLTAGE": 4,
        "POWER_SUPPLY_HEALTH_UNSPEC_FAILURE": 5,
        "POWER_SUPPLY_HEALTH_COLD": 6,
        "POWER_SUPPLY_HEALTH_WATCHDOG_TIMER_EXPIRE": 7,
        "POWER_SUPPLY_HEALTH_SAFETY_TIMER_EXPIRE": 8,
        "POWER_SUPPLY_TECHNOLOGY_UNKNOWN": 0,
        "POWER_SUPPLY_TECHNOLOGY_NIMH": 1,
        "POWER_SUPPLY_TECHNOLOGY_LION": 2,
        "POWER_SUPPLY_TECHNOLOGY_LIPO": 3,
        "POWER_SUPPLY_TECHNOLOGY_LIFE": 4,
        "POWER_SUPPLY_TECHNOLOGY_NICD": 5,
        "POWER_SUPPLY_TECHNOLOGY_LIMN": 6,
    },
    "msgtype": "sensor_msgs/msg/BatteryState"
}  
```

Why each part matters:

1) Enums make the timeseries data easier to interpret. Instead of seeing a raw number like 2, you can map it back to a meaningful label like POWER_SUPPLY_STATUS_DISCHARGING.

2) The original message type lets us correctly reconstruct the type when converting back from Mosaico to ROS. Previously this wasn't possible: a Mosaico `Pose` type could have originated from `sensor_msgs/msg/Pose`, `sensor_msgs/msg/PoseStamped`, `sensor_msgs/msg/PoseWithCovariance`, or `sensor_msgs/msg/PoseWithCovarianceStamped`, but it always defaulted to `sensor_msgs/msg/Pose`. Now the source type is preserved and used when translating back to ROS.

Fixes #568


mosaico-sdk-py/src/mosaicolabs/ros_bridge/loader.py
```diff
@@ -670,8 +672,16 @@ def msg_types(self) -> List[str | None]:
...
-        return self.streamer
+        return self
+
+    def __next__(self) -> tuple[str, dict[str, Any], Message]:
+
+        t_name, msg = self.streamer.__next__()
+
+        t_metadata = self.seq_handler.get_topic_handler(t_name).user_metadata
+
+        return t_name, t_metadata, msg
```
fdicorato: why are we returning the user metadata **at each (streaming) iteration step**? It can be a lot of data to return per each step

mosaico-sdk-py/src/mosaicolabs/ros_bridge/sequence_extractor.py
```diff
@@ -317,23 +319,31 @@ def _process_message(
...
-            ui.update_status(ms_topic, "No Adapter", style="yellow")
+            ui.update_status(t_name, "No Adapter", style="yellow")
             ui.advance_global()
             return
 
         # --- Translate Check ---
-        ros_msg = adapter.to_ros(ms_msg, self.typestore)
+        ros_msg_type = None
+        try:
+            ros_msg_type = t_metadata["_ros_"]["msgtype"]
```
fdicorato: I think I understand why you are returning the topic metadata at each message iteration. No, I would better design this part and avoid returning the topic metadata at each step. One approach can be asking the MosaicoLoader for the ROS msgtype corresponding to a certain topic (since it has access to the sequence handler and therefore to the related topic handlers

fdicorato (changes requested):
The extraction of the ROS msgtype from the topic metadata should be refactored

mosaico-sdk-py/src/mosaicolabs/examples/ros_injection/reconstruct_rosbags.py
```diff
@@ -0,0 +1,58 @@
...
+    "r2b_robotarm_0",
+]
+
+
+ROS_DISTRO = Stores.ROS2_JAZZY
+
+
+def main():
+
+    # --- PHASE 0: Custom messages registration ---
```
fdicorato: Comment not right

mosaico-sdk-py/src/mosaicolabs/examples/ros_injection/reconstruct_rosbags.py
```diff
@@ -0,0 +1,58 @@
...
+    for sequence in SEQUENCE_NAMES:
```
fdicorato, line 31: I think can be better to set the sequence to be converted, via a CLI param `sequence`, instead of converting everything.

mosaico-sdk-py/src/mosaicolabs/ros_bridge/sequence_extractor.py
```diff
@@ -317,23 +343,24 @@ def _process_message(
 
         # If no adapter can be found, ingore the topics from now on
         if adapter is None:
```
fdicorato, line 337: This should never happen, because not-adapted ontologies are checked during _resolve_sequence(). If you want to keep the check for Noneness (get_default_mosaico_adapter can return None) for linter checks, you can return directly, because this is an impossible condition

mosaico-sdk-py/src/mosaicolabs/ros_bridge/sequence_extractor.py
```diff
@@ -317,23 +343,24 @@ def _process_message(
...
+            ui.update_status(t_name, "No Adapter", style="yellow")
             ui.advance_global()
             return
 
         # --- Translate Check ---
-        ros_msg = adapter.to_ros(ms_msg, self.typestore)
+        ros_msg_type = self.mosaico_loader.resolve_ros_msgtype(t_name)
+        ros_msg = adapter.to_ros(ms_msg, self.typestore, ros_msg_type)
 
         if not ros_msg:
```
fdicorato: to_ros raises if the conversion cannot be made. protect with a try,except if want to manage the erroneous conversion.
Still uncertain about what to do in this situation: blacklist the topic or trying with the next message?

mosaico-sdk-py/src/mosaicolabs/ros_bridge/sequence_extractor.py
```diff
@@ -317,23 +343,24 @@ def _process_message(
...
-            self.ignored_topics.add(ms_topic)
+            self.ignored_topics.add(t_name)
```
fdicorato, line 346: Maybe here can be useful to use a dedicated list, making explicit that this topic had errors during conversion (like `_encoding_error` or other). In this way you can (if needed) implement policies like
* blacklist topic at the first N errors
* ignore and go over
* ...
These policies can be set into config class

mosaico-sdk-py/src/mosaicolabs/ros_bridge/sequence_extractor.py
```diff
@@ -317,23 +343,24 @@ def _process_message(
...
         # If no adapter can be found, ingore the topics from now on
         if adapter is None:
-            self.ignored_topics.add(ms_topic)
+            self.ignored_topics.add(t_name)
             logger.warning(
-                f"Could not find Adapter for topic '{ms_topic}' of type '{mosaico_type}'. Skipping the topic associated to this message"
+                f"Could not find Adapter for topic '{t_name}' of type '{mosaico_type}'. Skipping the topic associated to this message"
             )
-            ui.update_status(ms_topic, "No Adapter", style="yellow")
+            ui.update_status(t_name, "No Adapter", style="yellow")
```
fdicorato: This UI setup should be done in other places (this portion of code should never be reached)

mosaico-sdk-py/src/mosaicolabs/ros_bridge/sequence_extractor.py
```diff
@@ -530,8 +566,8 @@ def ros_sequence_extractor():
 
     args = parser.parse_args()
 
-    selected_distro = Stores.__members__.get(args.ros_distro)
-    selected_storage_plugin = StoragePlugin.__members__.get(args.storage_plugin)
+    selected_distro = Stores.__members__[args.ros_distro]
+    selected_storage_plugin = StoragePlugin.__members__[args.storage_plugin]
```
fdicorato, line 565: here if ros_distro or storage_plugin are wrong, the code abruptly raises

fdicorato (changes requested):

mosaico-sdk-py/src/mosaicolabs/ros_bridge/sequence_extractor.py
pucciland95: All next messages related to that topic will have the same `ros_msg_type`. Therefore, I think it would be better to blacklist the topic itself since I expect always the same behavior for all future messages of the same topic. What do you think?

mosaico-sdk-py/src/mosaicolabs/ros_bridge/sequence_extractor.py
pucciland95: Yes, but this should be prevented by the `choices` flag there is in the `parser` (with a suggestion that the user provided input is wrong):

``` python
  # ROS arguments
  parser.add_argument(
      "--ros_distro",
      default="ROS2_JAZZY",
      choices=[s.name for s in Stores], # <-- This one here
      help="Target ROS distribution for messages. If not set defaults to ROS2_HUMBLE",
  )
  parser.add_argument(
      "--storage_plugin",
      default="MCAP",
      choices=[sp.name for sp in StoragePlugin], # <-- This one here
      help="Storage plugin to save rosbag. If not set defaults to MCAP",
  )
```

Do you think it still would be better to double check?

mosaico-sdk-py/src/mosaicolabs/examples/ros_injection/reconstruct_rosbags.py
pucciland95: As discussed, `cli.py` will be refactored in next PR so different examples can have different parameters

mosaico-sdk-py/src/mosaicolabs/ros_bridge/loader.py
```diff
@@ -78,18 +74,8 @@ def resolved_topics(self) -> List[str]:
...
-    @property
-    def filtered_topics(self) -> List[str]:
-        """This should return the user filtered topics thorugh the glob pattern of the loaded data as strings"""
-        ...
-
-    @property
-    def msg_types(self) -> List[str | None]:
-        """This should return the types of the loaded data as strings"""
+    def rejected_topics(self) -> List[Tuple[str, str, str]]:
+        """This should return a list of tuples containing all the rejected topics, the reason and the color (topic_name, rej_reason, color)"""
```
fdicorato: The color is just a matter of the UI progress bar aspect, not something related to the loader itself. For separation of responsibilities, I would avoid the loader to be responsible of setting the color. Maybe it can be possible to add a map into the UI class that associates a color for each `rej_reason`, and to configure the colors into the UI context, instead of the loader. For better design, it could be better to implement rejection reasons as an Enum

fdicorato (changes requested):

mosaico-sdk-py/src/mosaicolabs/ros_bridge/loader.py
pucciland95: Yes, makes sense

fdicorato (approved):

PR #578, opened 2026-06-16
feat(mosaicod): Enable complex column types queries

### LINKED ISSUE #558: Enable complex column types queries
To support complex data structures, the query syntax introduces explicit array index and quantification specifiers. This allows for precise filtering and matching criteria over nested multi-valued fields, such as Arrow list arrays stored within Parquet columns.

The existing ontology field dot-path notation is extended with an index specifier to define how predicates should be evaluated across elements of a list.

### Index Specifiers

Specifier | Semantics
-- | --
[i] | The element at position i must satisfy the predicate.
[?] | At least one element must satisfy the predicate.
[!] | Every element must satisfy the predicate.

Add syntax for addressing Parquet columns that store Arrow list arrays. The existing ontology field dot-path notation is extended with an index specifier.

#### Examples
```py
object.list[?] = 1        # true if any element equals 1

object.list[?] $gt 1        # true if any element is greater than 1

object.list[!] = 1        # true if all elements equal 1

object.list[!] $in [1,2,3]  # true if every element is one of {1, 2, 3}

object.list[?] $in [1,2,3]  # true if some element is one of {1, 2, 3}
```

### Supported Operators per Specifier

Operator | `[i]` | `[?]` any | `[!]` all | plain list
-- | -- | -- | -- | --
`=` | ✓ | ✓ | ✓ | element-wise
`!=` | ✓ | ✓ | ✓ | element-wise
`>` | ✓ | ✓ | ✓ | —
`<` | ✓ | ✓ | ✓ | —
`between` | ✓ | ✓ | ✓ | —
`in` | ✓ | ✓ | ✓ | —
`match` (strings only) | ✓ | ✓ | ✓ | —

### Ragged (Struct-of-Lists) Columns
Ragged types (e.g. `list[SomeStruct]`) cannot be queried at the container or intermediate struct level. Only **leaf scalar fields** are addressable. The same `[i] / [?] / [!]` specifiers apply.</p>

LuigiTer:
Add support for querying Arrow list columns stored in Parquet files using the `[i]`, `[?]` and `[!]` index specifiers. This allows writing filter expressions that target individual elements, require at least one element to match, or require all elements to satisfy a predicate.

Replace the fragile value/tag_offset pair in `OntologyField` with a plain tag: String field alongside the existing `OntologyFieldPath`. The full string representation is now reconstructed on demand via Display, removing the risk of the offset going stale if the raw string were ever mutated.

`OntologyFieldPath` and `IndexSpecifier` gain `PartialEq, Eq, Hash` and `Display` impls so that `OntologyField` can derive its own equality and hashing from the structured data rather than from a cached string.

Close #558

### DIFF
--- a/mosaicod/crates/mosaicod-core/src/types/chunk.rs
+++ b/mosaicod/crates/mosaicod-core/src/types/chunk.rs
@@ -22,14 +22,13 @@ pub enum Stats {
     Numeric(NumericStats),
     Textual(TextualStats),
     Unsupported,
+    List(NumericStats),
+    ListTextual(TextualStats),
 }
 
 impl Stats {
     pub fn is_unsupported(&self) -> bool {
-        if let Stats::Unsupported = self {
-            return true;
-        }
-        false
+        matches!(self, Stats::Unsupported)
     }
 }
 
--- a/mosaicod/crates/mosaicod-core/src/types/chunk.rs
+++ b/mosaicod/crates/mosaicod-core/src/types/chunk.rs
@@ -22,7 +22,7 @@ pub enum Stats {
     Numeric(NumericStats),
     Textual(TextualStats),
     Unsupported,
-    List(NumericStats),
+    ListNumeric(NumericStats),
     ListTextual(TextualStats),
 }
 
--- a/mosaicod/crates/mosaicod-ext/src/arrow.rs
+++ b/mosaicod/crates/mosaicod-ext/src/arrow.rs
@@ -1,4 +1,6 @@
-use arrow::array::{ArrayRef, AsArray, RecordBatch, StructArray};
+use arrow::array::{
+    ArrayRef, AsArray, FixedSizeListArray, LargeListArray, ListArray, RecordBatch, StructArray,
+};
 use arrow::datatypes::{DataType, Field, FieldRef, Schema, SchemaRef};
 use mosaicod_core::{self as core, params, types};
 use parquet::arrow::async_reader::{AsyncFileReader, ParquetObjectReader};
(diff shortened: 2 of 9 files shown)


LuigiTer:
The current changes do not add support for list-of-struct columns. When a schema contains a `List<Struct<...>>` field, the backend fails at write time with a cast error.

The failure path is:
1. `SchemaFlattenerIter` emits the list column as a single leaf (e.g. "poses")
2. `stats_from_arrow_field` assigns it `Stats::List(NumericStats)` since the element type is not textual
3. During stats collection, `list_child_values` extracts the inner StructArray, then `cast_array_to_numeric` fails because Struct is not a numeric type
4. The error propagates to the client

This will be solved in a different issue.

mosaicod/crates/mosaicod-core/src/types/chunk.rs
```diff
@@ -22,14 +22,13 @@ pub enum Stats {
     Numeric(NumericStats),
     Textual(TextualStats),
     Unsupported,
+    List(NumericStats),
```
cabba:
```suggestion
    ListNumeric(NumericStats),
```

mosaicod/crates/mosaicod-ext/src/arrow.rs
```diff
@@ -298,10 +300,41 @@ pub fn stats_from_arrow_field(field: &Field) -> types::Stats {
...
         dt if is_numeric(dt) => Stats::Numeric(NumericStats::new()),
         dt if is_textual(dt) => Stats::Textual(TextualStats::new()),
+        DataType::List(elem) | DataType::LargeList(elem) => {
+            if is_textual(elem.data_type()) {
+                Stats::ListTextual(TextualStats::new())
+            } else {
+                Stats::List(NumericStats::new())
+            }
+        }
+        // Do we need to handle the fixed size?
```
cabba: Is it a question or a statement?

mosaicod/crates/mosaicod-ext/src/arrow.rs
```diff
@@ -298,10 +300,41 @@ pub fn stats_from_arrow_field(field: &Field) -> types::Stats {
...
+                Stats::List(NumericStats::new())
+            }
+        }
         _ => Stats::Unsupported,
     }
 }
 
+/// Extracts the flattened child values array from a list array.
+/// Returns None if the array is not a recognized list type.
+fn list_child_values(array: &ArrayRef) -> Option<ArrayRef> {
```
cabba:
```suggestion
fn list_child_values_from_array(array: &ArrayRef) -> Option<ArrayRef> {
```

mosaicod/crates/mosaicod-query/src/filter.rs
```diff
@@ -218,62 +218,166 @@ where
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum IndexSpecifier {
+    /// Access the element at a specific position: [0], [42].
+    At(usize),
+    /// At least one element must satisfy the predicate: [?].
+    Any,
```
cabba:
```suggestion
    AtLeastOne,
```

mosaicod/tests/tests/ontology_query.rs
```diff
@@ -1,11 +1,99 @@
...
+};
 use arrow::datatypes::{DataType, Field, Schema};
 use mosaicod_db as db;
 use serde_json::json;
 use std::sync::Arc;
 use tests::{actions, common};
 
+use arrow::array::ArrayRef;
+
+fn int_list_batch(ts_start: i64, values: &[i64], list_test: &[Vec<i64>]) -> RecordBatch {
```
cabba, line 15: Add doc. From the code is not super clear the purpose of the test

mosaicod/tests/tests/ontology_query.rs
```diff
@@ -1,11 +1,99 @@
...
+        vec![
+            Arc::new(Int64Array::from(timestamps)) as ArrayRef,
+            Arc::new(Int64Array::from(values.to_vec())) as ArrayRef,
+            Arc::new(list_array) as ArrayRef,
+        ],
+    )
+    .unwrap()
+}
+
+fn string_list_batch(ts_start: i64, values: &[i64], list_test: &[Vec<&str>]) -> RecordBatch {
```
cabba, line 61: Add doc. From the code is not super clear the purpose of the test

cabba (changes requested):

cabba (approved):

PR #579, opened 2026-06-16
perf: iterate dataclass fields in _to_dict instead of scanning dir()

facontidavide:
rosbags message objects are dataclasses with no __slots__, so _to_dict() fell back to scanning dir(message) -- enumerating and filtering the entire attribute space (methods, dunders, constants) on every nested object, recursively, for every message. Iterating __dataclass_fields__ instead is a hot-path win.

Measured on nav_msgs/Odometry (23,991 msgs): 61.9 us/msg -> 17.9 us/msg (~3.5x faster). Output is byte-identical (verified field-by-field), so no behavior change. Falls back to __slots__ (unchanged) and then dir() for non-dataclass objects.

We value everyone’s time and effort.  
To avoid work on unrequested or duplicate features, **pull requests must meet the requirements below**.

Please confirm all of the following before submitting this PR:

- I have read and understood the `CONTRIBUTING.md` file.
- I have opened a discussion related to this feature or bug.
- This work was previously discussed and agreed upon by maintainers.
- I confirm that this contribution does not include code (or comments) copied from proprietary sources or from open-source projects with incompatible licenses.

Now you can delete all the content above and submit your pull request. Thanks for the contribution!

### DIFF
--- a/mosaico-sdk-py/src/mosaicolabs/ros_bridge/helpers.py
+++ b/mosaico-sdk-py/src/mosaicolabs/ros_bridge/helpers.py
@@ -18,11 +18,18 @@ def _to_dict(message: Any) -> Any:
     """
     if hasattr(message, "__msgtype__"):
         data_dict = {}
-        fields = getattr(
-            message,
-            "__slots__",
-            [k for k in dir(message) if not k.startswith("_") and k != "__msgtype__"],
-        )
+        # rosbags messages are dataclasses without __slots__. Iterating the
+        # declared dataclass fields is ~3.5x faster than scanning dir(message)
+        # (which enumerates and filters the whole attribute space on every
+        # nested object) and yields identical data.
+        fields = getattr(message, "__slots__", None)
+        if fields is None:
+            dataclass_fields = getattr(message, "__dataclass_fields__", None)
+            if dataclass_fields is not None:
+                fields = list(dataclass_fields.keys())
+            else:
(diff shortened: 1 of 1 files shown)


mosaico-sdk-py/src/mosaicolabs/ros_bridge/helpers.py
```diff
@@ -18,11 +18,21 @@ def _to_dict(message: Any) -> Any:
...
+        # declared dataclass fields is ~3.5x faster than scanning dir(message)
+        # (which enumerates and filters the whole attribute space on every
+        # nested object) and yields identical data.
+        fields = getattr(message, "__slots__", None)
+        if fields is None:
+            dataclass_fields = getattr(message, "__dataclass_fields__", None)
+            if dataclass_fields is not None:
+                fields = list(dataclass_fields.keys())
+            else:
+                fields = [
```
fdicorato: Here I would simply do:
```python
fields = dir(message)
```
as the filtering of `fields` is made later (line 37)

fdicorato (changes requested):
Thank you! Please, check the proposed changes.

mosaico-sdk-py/src/mosaicolabs/ros_bridge/helpers.py
facontidavide: done

fdicorato (approved):
Thank you!

PR #581, opened 2026-06-18
feat(mosaicod): Enable complex query for list of structs

### LINKED ISSUE #580: Enable complex query for list of structs
Ontology query filters currently support scalar fields and flat list fields (e.g. `List<f64>`), but not lists whose elements are structs.

A query like:
```json
{ "ontology": { "robot.pose[?].acceleration.x": { "$gt": 3.0 } } }
```

where readings is of type `List<Struct<x: f64, y: f64>>` fails at runtime with an error.

Support for [?] and [!] specifiers on List<Struct> sub-fields should be added, along with the corresponding stats collection.

LuigiTer:
Added support to manage list of struct for complex query

Close #580

### DIFF
--- a/mosaicod/crates/mosaicod-ext/src/arrow.rs
+++ b/mosaicod/crates/mosaicod-ext/src/arrow.rs
@@ -81,7 +81,7 @@ pub fn empty_schema_ref() -> Arc<Schema> {
     Arc::new(Schema::empty())
 }
 
-/// Extract the schemna from a parquet reader object
+/// Extract the schema from a parquet reader object
 pub async fn schema_from_parquet_reader(
     reader: &mut ParquetObjectReader,
 ) -> Result<SchemaRef, Error> {
@@ -176,17 +176,23 @@ pub fn array_from_flat_field_name(
     }
 
     let top_level_name = subfields[0];
-    let mut current_array = batch.column_by_name(top_level_name).ok_or_else(|| {
-        arrow::error::ArrowError::SchemaError(format!(
-            "can't find top level field `{0}`",
-            top_level_name
-        ))
-    })?;
+    let mut current_array: ArrayRef =
+        Arc::clone(batch.column_by_name(top_level_name).ok_or_else(|| {
+            arrow::error::ArrowError::SchemaError(format!(
+                "can't find top level field `{0}`",
+                top_level_name
+            ))
+        })?);
 
     // Iterate and traverse the remaining nested path components
     //
     // *Note*: only structs fields are supported
     for subfield in &subfields[1..] {
+        // If the current array is a list, flatten it to reach the inner struct values.
+        if let Some(inner) = list_child_values_from_array(&current_array) {
+            current_array = inner;
+        }
+
         let struct_array = current_array
             .as_any()
             .downcast_ref::<StructArray>()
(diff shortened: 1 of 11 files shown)


mosaicod/crates/mosaicod-ext/src/arrow.rs
```diff
@@ -249,8 +255,29 @@ impl Iterator for SchemaFlattenerIter {
                     }
                     // Skip the struct itself and continue to process its children.
                 }
-
-                // 2) Primitive/Leaf Type - This is a field we want to yield.
+                // 2) List<Struct<...>> - recurse into the struct's children.
+                //    List<primitive> falls through to the leaf case below.
+                DataType::List(inner) | DataType::LargeList(inner) => {
+                    if let DataType::Struct(children) = inner.data_type() {
+                        for child_field in children.iter().rev() {
```
DevMattG, line 262: Why reverse iterating and push_front instead of forward iterating and push?

DevMattG (changes requested):

mosaicod/crates/mosaicod-ext/src/arrow.rs
LuigiTer: Looking at the existing code, the pattern is necessary to preserve the original field order in the DFS traversal.

DevMattG (approved):

mosaicod/crates/mosaicod-ext/src/arrow.rs
```diff
@@ -197,15 +203,15 @@ pub fn array_from_flat_field_name(
                 ))
             })?;
 
-        current_array = struct_array.column_by_name(subfield).ok_or_else(|| {
+        current_array = Arc::clone(struct_array.column_by_name(subfield).ok_or_else(|| {
```
cabba, line 206: Can we avoid the `Arc::clone` inside the for loop and cloning only at the end?

mosaicod/crates/mosaicod-query/src/timeseries.rs
```diff
@@ -365,7 +446,128 @@ fn all_op_to_df_expr<V: Into<Value>>(arr: Expr, op: Op<V>) -> Option<Expr> {
...
+
+/// Builds the predicate body for a lambda operating on a single struct element of a list.
+///
+/// When filtering `readings[?].x > 3`, DataFusion needs an expression of the form:
+///
+/// ```text
+/// array_any_match(col("readings"), lambda(["elem"], <body>))
+/// ```
+///
+/// This function produces `<body>`: it creates a `lambda_var("elem")` (the placeholder for
```
cabba: What is `<body>`?

mosaicod/crates/mosaicod-query/src/timeseries.rs
```diff
@@ -365,7 +446,128 @@ fn all_op_to_df_expr<V: Into<Value>>(arr: Expr, op: Op<V>) -> Option<Expr> {
...
+/// Builds the DataFusion expression for `[?]` on a `List<Struct<…>>` column, where the
+/// predicate targets a sub-field inside each struct element (e.g. `readings[?].x > 3`).
+fn any_op_struct_to_df_expr<V: Into<Value>>(
+    arr: Expr,
+    field_segments: &[String],
+    op: Op<V>,
+) -> Option<Expr> {
+    let body = struct_elem_predicate(field_segments, op)?;
+    Some(array_any_match(arr, lambda(["elem"], body)))
+}
```
cabba, line 574: We need to use a uniform naming:
- at least one
- any

If more readable (and preferred) we can use `any` but we need to change the enum value.
Mixing the two can lead to confusion.

cabba (approved):

PR #603, opened 2026-07-02
fix(py): replace wait_for_available with version-based probe for stability

### LINKED ISSUE #602: Avoid calling `list_flights` for every `MosaicoClient.connection`
Currently this piece of code 

```py
from mosaicolabs import MosaicoClient
client = MosaicoClient.connect("localhost", 6726)
```
calls the PyArrow's `wait_for_available` function before returning the client instance (timeout check). Such method calls `list_flights` internally to check for server health.

Since the `list_flight` can produce a long output we should avoid calling it at every connection.

fdicorato:
Fixes #602

### DIFF
--- a/mosaico-sdk-py/src/mosaicolabs/comm/connection.py
+++ b/mosaico-sdk-py/src/mosaicolabs/comm/connection.py
@@ -4,6 +4,8 @@ Connection Management Module.
 This module handles the creation and management of PyArrow Flight network connections.
 """
 
+import json
+import time
 from dataclasses import dataclass
 from enum import Enum
 from typing import Any, Optional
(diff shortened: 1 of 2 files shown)


mosaico-sdk-py/src/mosaicolabs/comm/connection.py
```diff
@@ -22,6 +24,29 @@
 logger = get_logger(__name__)
 
 
+def _wait_for_available(client: fl.FlightClient, timeout: int) -> None:
+    """Probe the server with a VERSION action until it responds or the timeout expires.
+
+    Uses do_action instead of list_flights to avoid triggering unintended
+    server-side list_flights handlers during connection setup.
```
cabba: This comment is a bit out of context for an external reader. 
Who calls `list_flights` in the first place?

I suggest to remove the comment or to describe better the context.

cabba (changes requested):

cabba (approved):

PR #604, opened 2026-07-02
refactor!: include_timestamp_range option removed from query request

### LINKED ISSUE #596: Remove include_timestamp_range flag and redundant timestamp payload from topic queries
### Problem
Currently, when making a query request  with the flag `include_timestamp_range=true`, the API appends the timestamp range data to every single retrieved topic in the sequence response payload.

This behavior introduces unnecessary payload bloat and redundancy. Individual topic metadata doesn't need to eagerly return these timestamp ranges here, as they can be fetched far more efficiently and precisely on-demand through `topic_filter_clusterize` query.

### Proposed Changes

* Remove the `include_timestamp_range` query parameter from the API.

* Clean up the response payloads removing the timestamp ranges.

DevMattG:
Closes #596


cabba (approved):
:+1:

PR #633, opened 2026-07-10
feat(sdk-py): add Unmodeled ontology support and class-free queries via Queryable Fields

### LINKED ISSUE #605: feat(py): Introduce Unmodeled ontology framework for schema-only (non-modeled) data ingestion
The SDK currently requires every ingestible data schema (ontology) to be backed by a hand-written `Serializable` Python class (e.g. `IMU`). The mosaico backend itself only needs data expressed as an Arrow schema. This restriction blocks ingestion of data types that have no corresponding SDK class, most notably ROS message types in the ros_bridge that haven't been explicitly adapted.

This change introduces the `Unmodeled` ontology base class, which lets the SDK dynamically generate a fully serializable, queryable ontology class from a raw pyarrow schema at runtime, with no Python class definition required.

Changes:

* New `Unmodeled` base class storing the payload as a generic `raw_data: Dict[str, Any]` instead of typed fields, plus `make_unmodeled_ontology_class(...)`, a factory that dynamically subclasses `Unmodeled` around an arbitrary `pa.StructType`.
* New `get_or_make_ontology_class` helper: resolves an ontology tag to its registered class if one exists, otherwise builds and registers an `Unmodeled` subclass on the fly from a supplied pyarrow schema.
* `Serializable`/`Message` factory methods `_encode`/`_decode`, unifying encoding across modeled and unmodeled classes.

fdicorato:
## Summary
- Introduce the `Unmodeled` ontology framework (`Unmodeled`, `make_unmodeled_ontology_class`,
  `resolve_ontology_class`) so the SDK can serialize, ingest, and query data whose schema is
  only known at runtime (e.g. ROS bag messages with no adapted SDK class), instead of requiring
  every ontology to be a hand-authored Python class.
- Split ontology identity into `__ontology_tag__` (the platform-facing tag, always stable) and
  `__registry_key__` (the SDK-local class registry key), so that multiple schema variants of the
  same message type sharing one ontology tag stay discoverable server-side under one consistent
  tag, while still avoiding local class-registration collisions. Also fixes a crash on schema-less
  ontologies and a check-then-create race when concurrently resolving a not-yet-registered tag,
  and adds `raw_data` validation against the declared schema at construction time.
- Refactor automatic ontology tag derivation to use the class name directly instead of converting
  it to snake_case, for consistent casing across the SDK (`IMU`, `GyroRaw`, etc.).
- Add class-free querying via the new `queryable_fields` module (`QueryableNumeric`,
  `QueryableString`, `QueryableBool`): build `.Q`-proxy-equivalent filter expressions directly
  from a `"tag.field.path"` string, including index syntax for querying inside list fields
  (`field[i]`, `field[!]`, `field[?]`) - useful whenever no resolved class is available to build
  a `.Q` proxy from, and for reaching into list fields that `.Q` doesn't support at all.
- Add full unit and integration test coverage for schema-variant resolution and class-free
  queries, refactor integration tests to use an injected client fixture, and add warnings for a
  deprecated classmethod plus consistent client closure across tests.
- Document all of the above: a new "Advanced: Ingesting Unmodeled Ontologies" section and a
  "Class-Free Queries" section (including list-field querying) in the Python SDK docs, a new
  "Query Unmodeled Ontologies" how-to guide in the main docs site, and full API reference coverage
  for the new public classes.

## Test plan
- [x] `./scripts/tests` passes
- [x] `mkdocs build --strict` builds the Python SDK docs cleanly

closes #605


mosaico-sdk-py/src/mosaicolabs/models/core/serializable.py
```diff
@@ -42,6 +42,19 @@
...
+    """
+    Computes a short, deterministic fingerprint identifying a pyarrow struct schema.
+
+    Two structurally identical schemas (same field names, types, nesting and
+    nullability) always produce the same fingerprint, regardless of when or where
+    they were built. Used to detect when a single ontology tag ends up associated
+    with more than one schema shape within the same process (e.g. two ontology
+    versions of the same dynamically-resolved, unmodeled data type).
+    """
+    return hashlib.sha1(str(struct).encode("utf-8")).hexdigest()[:10]
```
pucciland95: Is there a reason why only the last 10 chars are taken from the hash?

fdicorato: I intentionally truncate the SHA-1 to the first 10 hexadecimal characters (40 bits) to keep the schema ID compact. Given the very small number of expected schema versions, the probability of an accidental collision is negligible, making the shorter identifier a reasonable trade-off

mosaico-sdk-py/src/mosaicolabs/models/core/unmodeled.py
```diff
@@ -0,0 +1,262 @@
...
+        else:
+            errors.extend(_diff_against_struct(value, field.type, f"{path}{name}."))
+
+    return errors
+
+
+class Unmodeled(
+    Serializable,
+    skip_schema_generation=True,
+    skip_query_proxy_ingestion=True,
```
pucciland95, line 83: It is not clear whether `skip_query_proxy_ingestion` should be `True` or `False` by default. Indeed, `make_unmodeled_ontology_class` docstring states:

```
`skip_query_proxy_ingestion` is left at its default (`False`), so the
`.Q` query proxy is still generated from `pyarrow_schema`.
```

However, here the the default is `True`.

Furthermore, at the end of `make_unmodeled_ontology_class` when the funciton dynamically creates the new subclass based on `Unmodelled` it does not set `skip_query_proxy_ingestion=False`, avoiding `.Q` injection for all `Unmodelled` subclasses making them impossible to query. Am I missing something?

Shouldn't this be:
```suggestion
    skip_query_proxy_ingestion=False,
```

fdicorato: The default value the docstring refers to is the default value of the parameter in `Serializable.__init_subclass__(..., skip_query_proxy_ingestion: bool = False)` (file mosaico-sdk-py/src/mosaicolabs/models/core/serializable.py:137).

The `skip_query_proxy_ingestion=True` in `Unmodeled` definition is relative to `Unmodeled` only; all other `Unmodeled` subclasses will inherit the default value set in `Serializable` class, if they do not set it otherwise. `Unmodeled` class will not have a Q proxy; all classes defined via `make_unmodeled_ontology_class` will. To test it, you can try this test:

```python
import pyarrow as pa
from mosaicolabs.enum.serialization_format import SerializationFormat
from mosaicolabs.models.core.unmodeled import Unmodeled, make_unmodeled_ontology_class

UnmodeledSubClass = make_unmodeled_ontology_class(
    "UnmodeledSubClass",
    None,
    SerializationFormat.Default,
    pa.struct(
        [
            pa.field("field", pa.float32()),
        ]
    ),
)

assert hasattr(UnmodeledSubClass, "Q")
assert not hasattr(Unmodeled, "Q")
```

fdicorato: Note: i've added the test `test_unmodeled_subclass_has_Q_proxy` in testing/unit/test_unmodeled.py, which distinctively runs the test above

mosaico-sdk-py/src/mosaicolabs/models/core/serializable.py
```diff
@@ -104,35 +170,53 @@ def __pydantic_init_subclass__(cls, **kwargs):
...
+            name: encode_to_dict(value) for name, value in self.model_dump().items()
+        }
 
     # --- Factory Methods ---
 
     @classmethod
-    def _create(cls, tag: str, *args, **kwargs) -> "Serializable":
+    def _decode(cls, *args, **kwargs) -> "Serializable":
         """
         Factory method to instantiate a specific ontology object by its tag.
```
pucciland95: Tag argument should be dropped from docstring since it has been removed

mosaico-sdk-py/src/mosaicolabs/models/query/generation/api.py
```diff
@@ -318,17 +321,75 @@ def all(self):
...
+                    field_type = val
+                    if _pyarrow_to_queryable(field_type) is not _QueryableUnsupported:
+                        result.append(key)
+                    elif isinstance(val, pa.DataType):  # (pa.DataType, expr_cls)
+                        field_type = val
+                        if (
+                            _pyarrow_to_queryable(field_type)
+                            is not _QueryableUnsupported
+                        ):
+                            result.append(key)
```
pucciland95: Copy paste error? There is the same condition `isinstance(val, pa.DataType)` within another `isinstance(val, pa.DataType)` and they both do the same actions

fdicorato: Yes, thank you!

mosaico-sdk-py/src/mosaicolabs/models/query/generation/mixins.py
```diff
@@ -261,47 +531,118 @@ def match(self, value: Any):
     def in_(self, *values: Any):
         return getattr(self, "_in")(*values, allowed_types=None)
```
pucciland95, line 559: Docstring missing for `in_` and `match` from _QueryableDynamicValue

mosaico-sdk-py/src/mosaicolabs/handlers/topic_handler.py
```diff
@@ -51,6 +53,7 @@ def __init__(
         client: fl.FlightClient,
         topic_model: Topic,
         ticket: fl.Ticket,
+        pyarrow_schema: pa.StructType,
```
pucciland95, line 56: Add arg documentation to docstring

docs/py/docs/SDK/query.md
```diff
@@ -558,6 +558,118 @@ with MosaicoClient.connect("localhost", 6726) as client:
...
+
+| Class | Supported Operators | Value Type |
+| --- | --- | --- |
+| [`QueryableNumeric`][mosaicolabs.models.query.queryable_fields.QueryableNumeric] | `.eq()`, `.neq()`, `.lt()`, `.leq()`, `.gt()`, `.geq()`, `.in_()`, `.between()` | `int`, `float` |
+| [`QueryableString`][mosaicolabs.models.query.queryable_fields.QueryableString] | `.eq()`, `.match()`, `.lt()`, `.leq()`, `.gt()`, `.geq()`, `.in_()` | `str` |
+| [`QueryableBool`][mosaicolabs.models.query.queryable_fields.QueryableBool] | `.eq()` | `bool` |
+
+#### Querying List Fields
+
+The `.Q` proxy skips list fields entirely (see [Unsupported Types (Lists and Tuples)](#supported-vs-unsupported-types) above) - there's no way to write `IMU.Q.covariance[0]`. `Queryable*` fields close this gap for *any* ontology, modeled or unmodeled, by extending the path syntax with an index selector that reaches inside a list.
```
pucciland95: This should be updated since now it is possible to query lists using the query proxy `.Q`

pucciland95 (changes requested):
All seems fine, just few comments on docstring and documentation

mosaico-sdk-py/src/mosaicolabs/models/query/generation/api.py
```diff
@@ -316,19 +281,90 @@ def all(self):
...
+            for key, val in self.__schema__.items():
+                if isinstance(
+                    val, (dict, *SUPPORTED_PYARROW_LIST_TYPES)
+                ):  # nested struct, _QueryableList or pa.ListType or pa.LargeListType or pa.FixedSizeListType
                     result.append(key)
+                elif isinstance(val, pa.DataType):
+                    if _pyarrow_to_queryable_mixin(val) is not _QueryableUnsupported:
+                        result.append(key)
+        else:
+            result.append(self.__schema__)
```
pucciland95, line 304: Since, in general, `self.__schema__` is either a `dict` or a `pa.Datatype` and here we excluded it to be a `dict` shouldn't we check to **not** be `_QueryableUnsupported` before appending to `result`?

pucciland95 (changes requested):
Just one last doubt. The rest is fine

mosaico-sdk-py/src/mosaicolabs/models/query/generation/api.py
fdicorato: Yes!

pucciland95 (approved):

PR #637, opened 2026-07-14
feat(py): ros adapter checks msgtype in topic metadata first and fallabacks to ontology tag

### LINKED ISSUE #597: Mosaico to ROS adapter always goes to default one
Mapping Mosaico ontologies to ROS messages always defaults to specified default Mosaico Adapter ignoring original `ros_msg_type` tha may be present within topic metadata.

The order when going Mosaico->ROS should be:

1) resolve Mosaico adapter through `ros_msg_type` in topic metadata using `_default_adapters` in `ROSBridge`
2) if 1) fails, resolve default adapter through `ontology_tag` and `_default_mosaico_adapters` in `ROSBridge`

pucciland95:
Fixes #597 and integrates changes to read topic message count from `do_get`

### DIFF
--- a/mosaico-sdk-py/src/mosaicolabs/handlers/topic_reader.py
+++ b/mosaico-sdk-py/src/mosaicolabs/handlers/topic_reader.py
@@ -141,11 +141,13 @@ class TopicDataStreamer:
             _decode_schema_metadata(reader.schema.metadata)
         )
         ontology_tag = topic_mdata.properties.ontology_tag
+        msg_count = topic_mdata.properties.msg_count
 
         rdstate = _TopicReadState(
             topic_name=topic_name,
             reader=reader,
             ontology_tag=ontology_tag,
+            msg_count=msg_count,
         )
         return cls(
             client=client,
@@ -283,6 +285,16 @@ class TopicDataStreamer:
             The ontology tag.
         """
         return self._rdstate.ontology_tag
+
+    @property
+    def msg_count(self) -> Optional[int]:
+        """
+        The number of messages associated with this streamer.
+
+        Returns:
+            The number of messages. None if an error occurred during message count retrieval
+        """
+        return self._rdstate.msg_count
 
     def __iter__(self) -> "TopicDataStreamer":
         """Returns self as iterator."""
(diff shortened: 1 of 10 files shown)


mosaico-sdk-py/src/mosaicolabs/ros_bridge/loader.py
```diff
@@ -689,20 +705,89 @@ def __init__(
...
+
+        adapter = None
+        resolved_rosmsg_type = None
+
+        # Try looking for adapter using msgtype (may still not be adapted)
+        if declared_rosmsg_type is not None:
+            adapter = ROSBridge.get_default_adapter(declared_rosmsg_type)
+            resolved_rosmsg_type = declared_rosmsg_type
+
+        # If here adapter has not been found and therefore declared_rosmsg_type
```
fdicorato: Just a detail: the code reaches this point in any case (i.e. if an adapter has been found also)., so 'if here' the adapter can be either None or not None. Please, fix the comment.

mosaico-sdk-py/src/mosaicolabs/ros_bridge/loader.py
fdicorato, line 798: Please, update the steps in the docstring: highlight the operations of extracting the adapters

mosaico-sdk-py/src/mosaicolabs/handlers/topic_reader.py
fdicorato, line 1: Please, add a `msg_count` property in the `SequenceDataStreamer` class also: the property will add the `msg_count` property per each topic and return the total number of messages.

Add some integration tests for validating the properties `SequenceDataStreamer/TopicDataStreamer.msg_count`. You can implement the tests in the file: 'testing/integration/test_endpoints.py'; you can test the total number of messages returned per each topic (via `TopicDataStreamer.msg_count`) and for the entire sequence (via `SequenceDataStreamer.msg_count`) and compare them to the actual number of messages stored in the `synthetic_sequence_data_stream` object

fdicorato (changes requested):

fdicorato (approved):

PR #639, opened 2026-07-15
feat(mosaicod): split cleanup and server into separate subcommands

### LINKED ISSUE #638: Split cleanup and server into separate subcommands
Currently `mosaicod` exposes a single run subcommand that both starts the gRPC server and spawns the cleanup background routine (see `Server::start_and_wait` in `crates/mosaicod-grpc/src/server.rs`). 
Cleanup is configured exclusively through environment params (`cleanup_time_interval`, `cleanup_retention_duration` in `crates/mosaicod-core/src/params.rs`).

We want to decouple these two services so that each one has its own subcommand and can be operated and deployed independently.

## Proposed changes

1. Rename **run** in **server**

- The run subcommand is renamed to server.
- It keeps all the arguments it already has (`--host`, `--port`, `--tls`, `--gzip`, `--api-key`) and the same behavior, with the sole exception that it no longer starts the cleanup routine.

2. New cleanup subcommand

- A new cleanup subcommand is added, dedicated exclusively to running the cleanup routine.
- It exposes two CLI arguments:
  - `--time-interval` : minimum interval (in seconds) between two consecutive cleanup runs. Default: 0.
  - `--retention-duration`: period (in seconds) an obsolete file is kept in the store before being permanently deleted. Default: 1 day.
- Semantics of --time-interval:
  - 0 -> run a single cleanup and then terminate (one-shot mode).
  - \> 0 -> run the recurring cleanup loop, sleeping `time_interval` seconds between runs (current behavior).

3. Remove cleanup params

- Remove `cleanup_time_interval` and `cleanup_retention_duration` from **Params** (and their `MOSAICOD_CLEANUP_TIME_INTERVAL` / `MOSAICOD_CLEANUP_RETENTION_DURATION` env variables). These values now come from the cleanup subcommand arguments.

## Notes

- Today `Cleanup::try_cleanup` returns when `time_interval.is_zero()`, and `Cleanup::run` loops forever using `time_interval` as the sleep duration. With the new semantics time_interval = 0 must mean "run exactly once and exit", so this early-return guard and the loop logic need to be revisited.
- The cleanup routine must no longer be spawned from `Server::start_and_wait`; that wiring should be removed from the server path and moved into the new cleanup subcommand entry point.
- Update the CLI Commands enum in `bin/mosaicod/src/main.rs` and the command modules under `bin/mosaicod/src/command/`.
- Update any documentation, deployment manifests, and docker/compose configs that reference the run subcommand or the removed env variables.

LuigiTer:
Rename the run subcommand to server and move the
cleanup routine to a dedicated cleanup subcommand with --time-interval and --retention-duration flags (default 0 and 86400).

- time_interval = 0 runs a single cleanup and exits
- cleanup is no longer spawned by the server nor configured via the removed MOSAICOD_CLEANUP_TIME_INTERVAL/RETENTION_DURATION env vars
- update docs and scripts

Close #638


mosaicod/crates/mosaicod-task/src/cleanup.rs
```diff
@@ -114,6 +118,12 @@ impl Cleanup {
                 }
             }
 
+            // perform a single cleanup and terminate.
+            if self.time_interval.is_zero() {
+                info!("Exiting cleanup routine. One-shot cleanup completed.");
+                break;
+            }
+
             // match self.can_start().await {
```
DevMattG: Remove commented code please

DevMattG (changes requested):
Considering the case of a single-shot run, should we skip the retention duration (even if explicitly set by the user) and delete the files right away? What would be best for the user? @LuigiTer @cabba

cabba (approved):

DevMattG (approved):

PR #644, opened 2026-07-16
feat(py): implement Mosaico<->ROS adapter for unmodeled ontologies

### LINKED ISSUE #641: Implement UnmodeledAdapter to encode/decode not adapted ros types
Currently `ros_bridge` discards ROS messages that does not have an implemented Mosaico Adapter. 
With the new addition of `Unmodeled` ontologies they can be translated to Mosaico and again back to ROS. 

The necessary steps should be:
- Extract `msgdef` from ros message and use it to deduce its PyArrow schema.
- Create a new `Unmodeled` ontology with the computed PyArrow schema.
- Create on-the-fly and register the new `UnmodeledAdapter` holding the new `unmodeled` ontology.
- Save within `UnmodeledAdapter` metadata `msgdef` to set later the Mosaico Topic metadata.
- Extract correct adapter from `ROSLoader` during injestion.

Remember to:
- [x] Implement `to_dict` of  `UnmodeledAdapter` 
- [x] Implement `to_ros` of  `UnmodeledAdapter` 
- [ ] Updated `ros_bridge` documentation.

> Note: Sequences with Unmodelled ontologies coming from something that is not a `rosbag` (`.hd5`, `.csv`, ...) cannot be translated back to ROS. This will probably handled in a different issue.

pucciland95:
Fixes #641


docs/main/docs/examples/reconstruct_rosbag.mdx
```diff
@@ -0,0 +1,100 @@
...
+        description="Cumulative counts from the right wheel encoder."
+    )
+    encoder_timestamp: MosaicoType.uint64 = MosaicoField(
+        description="Timestamp of the encoder ticks."
+    )
+```
+
+## No Adapter Needed to Reconstruct
+
+The [ROS Ingestion](./ros_ingestion) example writes `EncoderTicksAdapter` by hand (imported here via `ros_injection.custom_ontology`, see [ROS Ingestion](./ros_ingestion#ros-adapter)) to get a fully-typed, queryable `EncoderTicks` model, but that step is a choice, not a requirement. Writing a reverse adapter is optional too: for any topic that reaches the extractor with no adapter registered for its ROS message type, the bridge falls back to the same automatic **Unmodeled Adapter** used at ingestion time. Since the original `.msg` definition was already captured as topic metadata when the sequence was ingested, the extractor can recover it, re-register it with the local ROS typestore, and reconstruct the native ROS message from it. See [Advanced: Ingesting Unmodeled Ontologies](https://docs.mosaico.dev/python-sdk/SDK/ontology/#advanced-ingesting-unmodeled-ontologies) for the full mechanics of this round-trip.
```
fdicorato: add a reference to 'Unmodeled Adapter used at ingestion time'

docs/main/docs/examples/ros_ingestion.mdx
```diff
@@ -90,6 +92,12 @@ class EncoderTicksAdapter(ROSAdapterBase[EncoderTicks]):
 ```
 
 
+## Writing an Adapter Is Optional
+
+`EncoderTicks` is the only message type in the R2B dataset without built-in ontology coverage, and above we gave it a hand-written adapter to get a fully-typed, queryable `EncoderTicks` model. That step is a choice, not a requirement.
+
+Even if `EncoderTicksAdapter` didn't exist, `RosbagInjector` would still ingest every `isaac_ros_nova_interfaces/msg/EncoderTicks` message without error: the bridge synthesizes an **Unmodeled Adapter** for it at runtime, converting the message's schema, embedded directly in the bag, into an equivalent Mosaico `Unmodeled` ontology class on the fly. The data is still ingested and fully queryable, just without the dedicated `EncoderTicks` Python class. See [Advanced: Ingesting Unmodeled Ontologies](https://docs.mosaico.dev/python-sdk/SDK/ontology/#advanced-ingesting-unmodeled-ontologies) for the full mechanics.
```
fdicorato, line 99: Add a reference when mentioning 'Unmodeled Adapter' and '`Unmodeled` ontology'

docs/py/docs/SDK/API_reference/models/unmodeled.md
fdicorato, line 1: API reference for Unmodeled is already here: docs/py/docs/SDK/API_reference/models/base.md. It can be ok to rename the 'base.md' file to 'core.md', for better clarity

docs/py/mkdocs.yml
```diff
@@ -82,6 +82,7 @@ nav:
         - SDK/API_reference/models/platform.md
         - SDK/API_reference/models/sensors.md
         - SDK/API_reference/models/time.md
+        - SDK/API_reference/models/unmodeled.md
```
fdicorato: It is not necessary

mosaico-sdk-py/src/mosaicolabs/models/core/types.py
fdicorato, line 1: Why remapping `large_string` to `string` or `large_binary` to `binary`? They are handled by the library, see `class MosaicoType` in mosaicolabs/models/core/types.py

mosaico-sdk-py/src/mosaicolabs/protocols/ros2msg.py
fdicorato, line 1: Consider using throwaway placeholder (`_`) for unpacked values that are not used, e.g.:
`ros_type, default_value = content`-> `ros_type, _ = content` if `default_value` is never used

mosaico-sdk-py/src/mosaicolabs/ros_bridge/adapters/unmodeled.py
```diff
@@ -0,0 +1,218 @@
...
+
+            item_node_type = cast(Nodetype, item_node_type)  # just for typechecker
+
+            if item_node_type is Nodetype.BASE:
+                raw_data[field_name] = np.array(raw_data[field_name])
+
+            elif item_node_type is Nodetype.NAME and isinstance(item_content, str):
+                # Check that all elements of raw_data[field_name] are dict
+                if any(not isinstance(x, dict) for x in raw_data[field_name]):
+                    raise TypeError(
```
fdicorato, line 91: Is the `TypeError` message wrong?

mosaico-sdk-py/src/mosaicolabs/ros_bridge/adapters/unmodeled.py
```diff
@@ -0,0 +1,218 @@
...
+        """
+        # Resolve ROS message to translate Mosaico message to if not defined in input
+        resolved_rosmsg_type = ros_msg_type or cls.get_default_ros_msg()
+        if not cls.is_rosmsg_type_valid(resolved_rosmsg_type):
+            raise TypeError(
+                f"Adapter {cls.__name__} does not support {resolved_rosmsg_type}"
+            )
+
+        # Checking presence in typestore of requested message
+        if typestore.types.get(resolved_rosmsg_type) is None:
```
fdicorato: You can store this in a variable, to avoid search (here) and direct access later (line 180)

mosaico-sdk-py/src/mosaicolabs/ros_bridge/adapter_base.py
```diff
@@ -253,16 +256,18 @@ def schema_metadata(cls, typestore: Typestore, ros_msg_type: str) -> Optional[di
...
+        if msg_def:
+            enum_list, _ = msg_def
+            out_dict["enums"] = {name: val for name, _, val in enum_list}
 
-        # Extract ENUM associated to ros_msg_type and adding it to the out dict with the ros_msg_type
-        enum_list, _ = msg_def
-        out_dict = {"enums": {name: val for name, _, val in enum_list}}
-        out_dict.update({"msgtype": ros_msg_type})
+        out_dict["msgtype"] = ros_msg_type
+        out_dict["msgdef"] = kwargs.get("ros_msg_def")
```
fdicorato: It should be sufficient to store `out_dict["msgdef"]` inside the `if msg_def:` branch only: if the  ros_msg_type is coeherent and correct, typestore will certainly contain the class definition. Therefore the `ros_msg_def` parameter is unnecessary

mosaico-sdk-py/src/mosaicolabs/ros_bridge/loader.py
fdicorato, line 1: `ROSLoader._resolve_connections` must be updated, to make the internal `_typestore` to match the `AnyReader.typestore`, where all the msgtypes (ROS standard AND custom ones), are defined. To do so, add the following line after opening the reader (after `self._reader.open()` in `ROSLoader_resolve_connections`):
`self._typestore = self._reader.typestore`

mosaico-sdk-py/src/mosaicolabs/ros_bridge/injector.py
```diff
@@ -468,15 +473,18 @@ def _process_message(
 
         # Retrieve the writer from SequenceWriter local cache or create new one on server
         twriter = seq_writer.get_topic_writer(ros_msg.topic)
+
         # Should theoretically not be None if exists returned True
         if twriter is None:
+            # --- Schema metadata Resolution ---
+            schema_metadata = adapter.schema_metadata(
+                self._loader._typestore, ros_msg.msg_type, ros_msg_def=ros_msg.msg_def
```
fdicorato: As above, ros_msg_def should be unnecessary, if msg_type is registered in self._loader._typestore. To be sure this happens, it will be necessary to fix the `ROSLoader._typestore`, as stated in https://github.com/mosaico-labs/mosaico/pull/644/changes#r3639277022

mosaico-sdk-py/src/mosaicolabs/ros_bridge/loader.py
```diff
@@ -392,21 +410,106 @@ def _resolve_connections(self):
                 self._filtered_topics.update({conn.topic: filtered_topic_info})
                 continue
 
-            # 2) Mosaico-adapted topic
-            if topic_info.msgtype and ROSBridge.is_msgtype_adapted(topic_info.msgtype):
+            # 2) Filter topics without a Mosaico-adapted
```
fdicorato: Update the comment: non-adapted topics are not filtered

mosaico-sdk-py/src/mosaicolabs/ros_bridge/loader.py
```diff
@@ -392,21 +410,106 @@ def _resolve_connections(self):
...
+            adapter = self._get_or_create_adapter(topic_info)
+
+            if adapter:
                 self._accepted_topics.update({conn.topic: topic_info})
-                self._connections.append(conn)
             else:
                 logger.warning(
-                    f"Skipping topic {conn.topic}: not-adapted msgtype {topic_info.msgtype}."
+                    f"Topic {conn.topic}: not-adapted msgtype {topic_info.msgtype}."
```
fdicorato: Update the error string: it is not a non-adapted topic. When `_get_or_create_adapter` return `None`?

mosaico-sdk-py/src/mosaicolabs/ros_bridge/loader.py
```diff
@@ -392,21 +410,106 @@ def _resolve_connections(self):
...
                 )
                 self._not_adapted_topics.update({conn.topic: topic_info})
```
fdicorato: Update the name of the dict: non adapted topics also are accepted

mosaico-sdk-py/src/mosaicolabs/ros_bridge/loader.py
```diff
@@ -392,21 +410,106 @@ def _resolve_connections(self):
...
+        msgtype: str = topic_info.msgtype
+        msgdef: str = topic_info.msgdef.data
+        pyarrow_schema = convert_ros2msg(msgdef, msgtype)
+
+        # Create the ontology
+        unmodeled_ontology = resolve_ontology_class(
+            ontology_tag=_class_name_from_ros_msgtype(topic_info.msgtype),
+            schema=pyarrow_schema,
+            # FIXME: how to pass this? Via Config?
+            serialization_format=SerializationFormat.Default,
```
fdicorato: the serialization format here is hard-coded and always set to `Default`. It could be nice if the user might set the serialization format per topic (or per msg_type) in the configuration

mosaico-sdk-py/src/mosaicolabs/ros_bridge/loader.py
```diff
@@ -778,14 +880,60 @@ def _resolve_topic_adapter(
...
+                    f"Cannot create Unmodeled Adapter for topic {t_handler.name} since its metadata do not contain msgtype or msgdef"
+                )
+                return None, resolved_rosmsg_type
+
+            # Create the ontology
+            unmodeled_ontology = resolve_ontology_class(
+                ontology_tag=t_handler.ontology_tag,
+                schema=t_handler._arrow_schema,
+                # FIXME: how to pass this? Via Config?
+                serialization_format=SerializationFormat.Default,
```
fdicorato: Why setting by default the serialization format instead of using `thandler.serialization_format`?

mosaico-sdk-py/src/testing/integration/ros_bridge/test_mosaico_loader.py
fdicorato, line 1: Here I would add some tests of msg_def with nested structs, to check robustness

fdicorato (changes requested):

mosaico-sdk-py/src/mosaicolabs/ros_bridge/loader.py
pucciland95: _get_or_create_adapter() can return None when you are dealing with a custom message and the .bag file holds no information about the message structure (i.e. a `.db3` file holds no information on the message structure, leading to an empty PyArrow struct and therefore to the creation of a wrong Unmodeled Ontology). In that case no Adapter can be resolved

fdicorato: Looking at the code, the function returns None if `topic_info.msgtype = None`. In the case you mention (db3 bags) the user must register the message definitions via config

pucciland95: In the new pushed code there is this case that previously was not accounted for:

``` python
if not pyarrow_schema:
    logger.warning(
        f"Topic {topic_info.msgtype} does not contain any schema and cannot be turned as an Unmodeled"
    )
    return None
```

`pyarrow_schema` is empty in case of `.db3` files, therefore in case the user **forgets** to register the custom ros message the `_get_or_create_adapter() ` should return `None` or the following code would not work since the `Unmodeled` would be created with an empty `pyarrow_schema`.

What do you think about: `f"Topic {conn.topic}: unresolved adapter for msgtype {topic_info.msgtype}."` ?

fdicorato (approved):

PR #658, opened 2026-07-22
feat(mosaicod): created store optimizer routine

### LINKED ISSUE #606: Implement Parquet file compaction (optimization) routine to resolve "tiny files" issue
## Context & Problem

Currently, when `mosaicod` receives topic data, it dumps it directly into the object store, performing only a direct conversion from Arrow to Parquet format. While this approach maximizes upload speeds, it results in the creation of a large number of **tiny files**. Over time, this heavily degrades read performance due to object store metadata overhead.

## Proposed Solution

We need to introduce a background optimization routine that runs periodically to merge (compact) small Parquet files into larger, optimized files based on a target threshold (either a specific number of rows or a target file size).

---

### 1. Configuration (Environment Variables)

A new environment variable will control how frequently this routine executes:

* **`MOSAICOD_OPTIMIZATION_TIME_INTERVAL`**: Sets the execution frequency.
* *Default value:* `1d` (once per day).

### 2. Database Changes

To track which topics require optimization and handle concurrency between multiple `mosaicod` instances, we need to modify the schema.

#### New Table: `topic_optimization_t`

This table acts as a job queue and distributed lock mechanism.

| Field Name | Type | Description |
| --- | --- | --- |
| `topic_id` | PK/FK | Foreign key mapping to the `topic_t` table. |
| `opt_path_in_store` | String | The new object store path where optimized files are being staged. |
| `start_unix_tstamp_secs` | Timestamp | Set when a node picks up the job. Acts as a sentinel to prevent concurrent processing by other instances. |

#### Alterations to `topic_t`

* Add field: `optimization_end_unix_tstamp_secs` (Timestamp, nullable). This allows the routine to quickly identify topics that have not yet been optimized.

---

### 3. Optimization Routine Workflow

The background routine should implement the following step-by-step logic:

1. **Discovery:** Fetch all topics from `topic_t` where `optimization_completion_tstamp` is `NULL`. Insert these topics into `topic_optimization_t` if they are not already listed.
2. **Locking:** Claim the first available topic in `topic_optimization_t` where `start_timestamp` is `NULL`. Immediately set `start_timestamp` to the current time to lock the job and prevent other concurrent `mosaicod` instances from picking it up.
3. **Staging:** Create a new target folder in the object store and update `opt_path_in_store` to point to it.
4. **Compaction:** Read and merge the original topic data files, writing the newly compacted Parquet files into the staging folder.
5. **Commit & Cleanup:** Once compaction completes successfully, execute a database transaction to:
    * Update `topic_t`: Replace the old `path_in_store` with `opt_path_in_store` and set `optimization_completion_tstamp` to the completion time.
    * Update `chunk_t`, `column_chunk_textual_t`, `column_chunk_numeric_t`: Replace the old info related to the topic with the new ones.
    * Delete the corresponding record from `topic_optimization_t`.

DevMattG:
Closes #606


mosaicod/crates/mosaicod-task/src/store_optimizer.rs
```diff
@@ -0,0 +1,381 @@
...
+                path: chunk_path,
+                metadata: chunk_metadata,
+                ontology_stats: stats,
+            });
+        }
+
+        Ok(chunk_stats)
+    }
+
+    async fn optimize(&self) -> Result<u32> {
```
LuigiTer: `topic_start_optimization` sets `start_unix_tstamp` and commits it (line 308) before the heavy work in `optimize_topic` ( line 316). 
If anything between that commit and `topic_optimization_delete` (line 321) fails, an err, a crash, or a shutdown, the row stays in `topic_optimization_t` with `start_unix_tstamp` set.

From there it's a dead end:
- `topic_next_to_be_optimized` only picks rows where `start_unix_tstamp IS NULL` (topic_optimization_record.rs line 29)
- `topic_update_optimization_list` won't re-add it (ON CONFLICT DO NOTHING)
- nothing ever resets start_unix_tstamp back to NULL

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
```diff
@@ -0,0 +1,381 @@
...
+use mosaicod_store as store;
+use std::cmp::max;
+use std::sync::Arc;
+use tokio_util::sync::CancellationToken;
+use tracing::{debug, error, info};
+
+const DEFAULT_TIME_INTERVAL: u32 = 86400;
+
+// Default max size for merged files.
+pub const DEFAULT_MAX_OUTPUT_FILE_SIZE: usize = 256_000_000; // MBi
```
LuigiTer: Do you mean `MB`?

LuigiTer (changes requested):

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
DevMattG: Actually they are bytes, I'll fix it

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
DevMattG: I added a check for topics staying too long in the optimization list with `start_unix_tstamp` set

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
```diff
@@ -0,0 +1,405 @@
...
+                    topic_to_optimize_record.topic_id,
+                )
+                .await?;
+
+                let chunk_stats = self
+                    .optimize_topic(&topic_record, &opt_path_in_store)
+                    .await?;
+
+                // Remove topic from optimization list once processed.
+                db::topic_optimization_delete(
```
LuigiTer: This part is outside a db transaction, so here the new path isn't referenced in any tables. If a cleanup starts in this window of time, the new path isn't referenced and then it is marked to_delete.

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
```diff
@@ -0,0 +1,405 @@
...
+                tx.commit().await?;
+
+                let topic_record = db::topic_find_by_id(
+                    &mut self.db.connection(),
+                    topic_to_optimize_record.topic_id,
+                )
+                .await?;
+
+                let chunk_stats = self
+                    .optimize_topic(&topic_record, &opt_path_in_store)
```
LuigiTer: If only one topic gives problem, the routine of optimization will be interrupted also for the remaining topics. Is this the correct behavior ?

LuigiTer (changes requested):

mosaicod/tests/tests/store_optimizer.rs
```diff
@@ -0,0 +1,1154 @@
...
+use mosaicod_db as db;
+use mosaicod_ext as ext;
+use mosaicod_rw::ToProperties;
+use mosaicod_store as store;
+use tests::{self, actions, cleanup, common, store_optimizer};
+
+// ===========================================================================
+// Store optimization routine. Single server tests
+// ===========================================================================
+
```
LuigiTer, line 30: I suggest to add tests to check  the data after the optimization are still readable and are the same before the routine starts.

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
DevMattG: We can change this part and don't stop at the first failure. What do you think if instead of a single u32 we make optimize function return a list of topic locators that were correctly optimized and a list of topic locators for which the optimization failed?

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
DevMattG: Yes we need to put `db::topic_optimization_delete` and `db::topic_optimization_complete` under the same transaction, or swap the order of execution

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
```diff
@@ -0,0 +1,423 @@
...
+        topic_record: &db::TopicRecord,
+        output_path_in_store: &types::TopicPathInStore,
+    ) -> Result<Vec<SerializedChunk>> {
+        debug!(
+            "Store optimization for topic {} started",
+            topic_record.locator()
+        );
+
+        // Estimates the batch size considering to fill iteratively the target output file size by a percentage.
+        // If max average row size is zero or the size of a single row is greater than max_file_size, set batcb_size to 1.
```
LuigiTer: typo `batch_size`

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
```diff
@@ -0,0 +1,423 @@
...
+
+                    db::topic_optimization_complete(
+                        &mut tx,
+                        topic_to_optimize_record.topic_id,
+                        types::Timestamp::now().as_i64(),
+                        opt_path_in_store,
+                    )
+                    .await?;
+
+                    self.update_chunk_stats(&mut tx, &topic_record, chunk_stats)
```
LuigiTer: We should also update `chunks_number` and `total_bytes`, right?

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
```diff
@@ -0,0 +1,423 @@
...
+                        .await?;
+
+                    tx.commit().await?;
+
+                    res.completed.push(topic_record.locator());
+                } else {
+                    // If the topic optimization fails, remove it from the optimization list anyway.
```
LuigiTer, line 436: Here i would log in some way the error.

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
```diff
@@ -0,0 +1,423 @@
...
+                    let mut tx = self.db.transaction().await?;
+
+                    // Remove topic from optimization list once processed and update path_in_store inside topic record.
+                    // These operations are done within the same transaction to prevent a cleanup routine scam in between.
+                    db::topic_optimization_delete(
+                        &mut tx,
+                        topic_to_optimize_record.topic_id,
+                        types::allow_data_loss(),
+                    )
+                    .await?;
```
LuigiTer: This will stop the optimization of all topics in case of failure

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
```diff
@@ -0,0 +1,423 @@
...
+                    res.completed.push(topic_record.locator());
+                } else {
+                    // If the topic optimization fails, remove it from the optimization list anyway.
+                    // It will be re-added at the next execution.
+                    db::topic_optimization_delete(
+                        &mut self.db.connection(),
+                        topic_to_optimize_record.topic_id,
+                        types::allow_data_loss(),
+                    )
+                    .await?;
```
LuigiTer: This will stop the optimization of all topics in case of failure

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
```diff
@@ -0,0 +1,423 @@
...
+                    / max_avg_row_size as usize,
+                1,
+            )
+        };
+
+        // Configure the session settings for file compaction
+        let config = df::execution::config::SessionConfig::new().with_batch_size(batch_size);
+
+        let runtime = Arc::new(
+            df::execution::runtime_env::RuntimeEnvBuilder::new()
```
LuigiTer: This declaration in `TimeseriesEngine` is a little different. There we use a memory pool and a disk manager builder probably to avoid OOM on big quantity of data. I think we need to do the same here.

LuigiTer (changes requested):

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
DevMattG: Correct, I'll fix it

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
```diff
@@ -0,0 +1,483 @@
...
+            "stale topics deleted from optimization table: {}",
+            stale_deleted
+        );
+
+        // Scans the database to search for topics not yet optimized and to put them inside topic optimization table.
+        db::topic_update_optimization_list(&mut self.db.connection()).await?;
+
+        let mut res = OptimizationResult::default();
+
+        while let Some(acquired_topic) = self.acquire_next_topic().await? {
```
LuigiTer: In case of error, the optimization will stops for remaining topics

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
```diff
@@ -0,0 +1,483 @@
...
+                    let errors_list = opt_res
+                        .failed
+                        .iter()
+                        .map(|(locator, err)| format!("\t{}: {}", locator, err))
+                        .collect::<Vec<_>>()
+                        .join("\n");
+
+                    info!(errors_list);
+                }
+                Err(e) => error!("Store optimization routine failed: {}", e),
```
LuigiTer, line 481: Here I would print, if possible, how many topics on total topics have been optimized until the error

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
```diff
@@ -0,0 +1,483 @@
...
+                    &mut tx,
+                    topic_to_optimize_record.topic_id,
+                    types::Timestamp::now(),
+                    opt_path_in_store.clone(),
+                )
+                .await?;
+
+                tx.commit().await?;
+
+                let topic_record = db::topic_find_by_id(
```
LuigiTer: Why this is out of the transaction?

LuigiTer (changes requested):

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
DevMattG: what policy would you apply to handle the error case?

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
DevMattG: because there is no need to keep it inside it. In general I would put inside a transaction the least number of operations required. Do you see reason why this should stay within the transaciton?

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
DevMattG: Not sure if this would be a valuable info. Considering a situation with multiple optimization routines running in parallel this would be misleading

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
LuigiTer: The one that you already have implemented. You can simply match the result of the `acquire_next_topic()` and log the error and skip to the next topic

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
LuigiTer: In case of error of  `topic_find_by_id`, the `start_unix_tstamp` and `opt_path_in_store` are valued, so it will remains at least 7 days in the db instead in case of error in a transaction we have the roll back. What do you think?

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
DevMattG: We can't just log the error due to how the next topic to acquire is picked. If we not delete the topic from the list, `acquire_next_topic()` will try to get always the same topic (except maybe in case of multiple routines running in parallel).  Even merely deleting the topic from the optimization list if the acquisition fails doesn't take into account the possibility of another routine acquiring the topic right before the deletion.
To handle also this case, we need to delete the topic only if `start_unix_tstamp` is `NULL`. Do you agree?

Moreover, deleting the topic from the optimization list prevents other routines (already in the `while-let` loop) from trying to acquire it. But this could be identified as a rare case...

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
DevMattG: If we explicitly handle the error returned by `acquire_next_topic()` there would be no need for this. But I will put it within the transaction anyway to be more robust

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
DevMattG: Sorry I need to correct me. It's not even possible to delete the topic from the list because I don't know which topic I was trying to acquire while the error happened. To avoid implementing too intricate logic I would prefer to just break the optimization in this case and wait for the next run to take care of re-trying

LuigiTer (approved):

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
```diff
@@ -0,0 +1,479 @@
...
+                path: chunk_path,
+                metadata: chunk_metadata,
+                ontology_stats: stats,
+            });
+        }
+
+        let mut tx = self.db.transaction().await?;
+
+        // Remove topic from optimization list once processed and update path_in_store inside topic record.
+        // These operations are done within the same transaction to prevent a cleanup routine scam in between.
```
cabba:
```suggestion
        // These operations are done within the same transaction to prevent a cleanup routine scan in between.
```

mosaicod/crates/mosaicod-task/src/store_optimizer.rs
```diff
@@ -0,0 +1,479 @@
+//! This module provides the optimization routine that improves the efficiency of the files in the object store,
+//! merging small files into bigger ones.
+
```
cabba, line 30: I will include a general technical overview of the optimization routine to clarify its overall behavior.

There is a lot of code and we need to have a general grasp of what the optimization does without reading all the code.

cabba (changes requested):

cabba (approved):

PR #673, opened 2026-07-27
feat(py): Add 'outside' operator to queryable fields and update documentation

### LINKED ISSUE #672: feat(py): Add outside([a, b]) operator in SDK query expressions
The outside operator is enabled for `_QueryableComparable` and `_QueryableDynamicValue`.

fdicorato:
- Introduced the 'outside' operator for numeric and string fields in the queryable models, allowing users to filter records outside a specified range.
- Updated relevant documentation across models and query builders to reflect the new operator.
- Enhanced query tests to validate the functionality of the 'outside' operator (integration and unit tests).

Closes #672

### DIFF
--- a/mosaico-sdk-py/src/mosaicolabs/query/generation/mixins.py
+++ b/mosaico-sdk-py/src/mosaicolabs/query/generation/mixins.py
@@ -156,16 +156,16 @@ class _QueryableComparable:
         Args:
             *values (Any): The candidate values, passed either as separate
                 positional arguments (`in_(v1, v2)`) or as a single list or
-                tuple (`in_([v1, v2])`). Every value must share the same type,
-                matching one of the types in `__mixin_supported_types__` for
-                this field.
+                tuple (`in_([v1, v2])`).
+                Both must be either `float` or `int`.
+
 
         Returns:
             The atomic comparison expression (`$in`).
 
         Raises:
             ValueError: If no values are provided.
-            TypeError: If the values don't all share the same, supported type.
+            TypeError: If both the values don't all share the same, supported type.
         """
         return getattr(self, "_in")(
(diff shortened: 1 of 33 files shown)


mosaico-sdk-py/src/mosaicolabs/query/generation/mixins.py
```diff
@@ -457,6 +496,52 @@ def in_(self, *values: str) -> "_QueryExpression":
             *values, allowed_types=self.__mixin_supported_types__
         )
 
+    def between(self, *values: str) -> "_QueryExpression":
+        """
+        Matches records where the field's value falls within an inclusive range.
+        The operator corresponds numerically to test wether v1 <= v <= v2.
```
pucciland95: wether -> whether. I pinpointed this but there are other in the file

mosaico-sdk-py/src/mosaicolabs/query/generation/mixins.py
```diff
@@ -457,6 +496,52 @@ def in_(self, *values: str) -> "_QueryExpression":
...
+            TypeError: If any value isn't a `str`.
+        """
+        return getattr(self, "_between")(
+            *values, allowed_types=self.__mixin_supported_types__
+        )
+
+    def outside(self, *values: str) -> "_QueryExpression":
+        """
+        Matches records where the field's value falls outside an exclusive range.
+        The operator corresponds numerically to test wether v < v1 || v > v2.
```
pucciland95: Maybe we should highlight that, since we are dealing with strings, we are doing a lexicographic comparison instead of a numerical one. What do you think? (the same then applies for `between` and `in`)

pucciland95 (changes requested):
Just some comments on docstrings

pucciland95 (approved):

PR #676, opened 2026-07-28
perf(py): Optimize Message construction by removing redundant work

fdicorato:
- Removed `default_factory=set` from the `_self_model_keys` `PrivateAttr` in `Message`. Pydantic runs an `inspect.signature()` check on every default factory for every instance created, and for a builtin like `set` this falls back to slow text-signature parsing (accounting for ~92% of per-instance construction time). The factory value was also always immediately discarded, since `model_post_init` unconditionally overwrites it. Removing it cuts `Message(...)` construction time by >30x
- Cached `_message_model_fields()`, `_get_schema()`, and the envelope/payload field-collision check (extracted into `_check_envelope_collision()`) with `lru_cache`. These were being recomputed on every `Message` instantiation and every `RecordBatch` schema lookup, even though their result only depends on the ontology class involved, not on instance data.

Net effect: cheaper `Message` construction and Arrow schema generation on high-throughput ingestion paths (streaming sensor data, bag ingestion), with no behavior change.

### DIFF
--- a/mosaico-sdk-py/src/mosaicolabs/models/core/message.py
+++ b/mosaico-sdk-py/src/mosaicolabs/models/core/message.py
@@ -11,7 +11,7 @@ middleware-level metadata (like recording timestamp_ns).
 import warnings
 from collections import defaultdict
 from functools import lru_cache
-from typing import Any, Dict, Optional, Type, TypeVar, Union
+from typing import Any, Dict, Optional, Set, Type, TypeVar, Union
 
 import pandas as pd
 import pyarrow as pa
@@ -162,7 +162,7 @@ class Message(BaseModel):
         """
         super().model_post_init(context)
         self._self_model_keys = Message._message_model_fields()
-        Message._check_envelope_collision(type(self.data))
+        Message._check_envelope_collision(self.data.__msco_pyarrow_struct__)
 
     def ontology_type(self) -> Type[Serializable]:
         """Retrieves the class type of the ontology object stored in the `data` field."""
@@ -254,7 +254,8 @@ class Message(BaseModel):
         return cls(data=data_obj, **message_kwargs)
 
     @classmethod
-    def _message_exclude_fields(cls):
+    def _message_exclude_fields(cls) -> Set[str]:
+        """Message fields that need to be excluded when encoding to dict"""
         return {"data"}
 
     @classmethod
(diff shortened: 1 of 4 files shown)


mosaico-sdk-py/src/mosaicolabs/models/core/message.py
```diff
@@ -262,8 +254,36 @@ def _decode(
...
+        class-level PyArrow schemas, not on any instance data.
+        """
+        colliding_fields = set(cls.__msco_pyarrow_struct__.names) & set(
+            data_type.__msco_pyarrow_struct__.names
+        )
+        if colliding_fields:
+            raise ValueError(
+                f"Fields name collision detected between class '{data_type.__name__}' "
+                f"and Message envelope. Colliding fields: {colliding_fields}."
+            )
```
pucciland95, line 287: I see that the same function could be called within `_get_schema()`, however to be used there it is necessary to change the `_check_envelope_collision()` to accepting `pa.StructType` rather than `Type[Serializable]`:

```suggestion
    @classmethod
    @lru_cache(maxsize=None)
    def _check_envelope_collision(cls, data_schema: pa.StructType) -> None:
        """
        Verifies there is no field name collision between the Message envelope 
        schema and a given ontology data type's schema.

        Cached per (cls, data_type) since the result only depends on the
        class-level PyArrow schemas, not on any instance data.
        """
        colliding_fields = set(cls.__msco_pyarrow_struct__.names) & set(data_schema)
        if colliding_fields:
            raise ValueError(
                f"Fields name collision detected between schema '{data_schema}' " 
                f"and Message envelope. Colliding fields: {colliding_fields}."
            )
```

what do you think?

mosaico-sdk-py/src/mosaicolabs/models/core/message.py
```diff
@@ -262,8 +254,36 @@ def _decode(
         return cls(data=data_obj, **message_kwargs)
 
     @classmethod
+    def _message_exclude_fields(cls):
+        return {"data"}
```
pucciland95:
```suggestion
    def _message_exclude_fields(cls) -> set[str]:
        """ Message fields that need to be excluded when encoding to dict """
        return {"data"}
```

pucciland95 (changes requested):
I propose one slight change so to use the new `_check_envelope_collision` also iin `get_schema` and another to improve docstring and added return type

pucciland95 (approved):

PR #690, opened 2026-07-30
bug(py): correct `_to_dict`splitting values and const +  added typesd…

### LINKED ISSUE #685: bug(py):  Problems during ingestion of ROS1 bags
# 1) Time and Duration not present in ROS1 msgdef 
All the ROS1 msgdef do not contain types like `Time` and `Duration` within their message definition. Therefore, when creating `Typesdict` from `rosbag` library using `msgdef` and `msgtype` the resulting  `Typesdict` will not hold any information about `Duration` and `Time` composition. 

Consequently, when looking for these types within `Typesdict` during `msgdef` -> `pa.StructType` conversion of an unmodeled ontology, it will not be able to transform fields like `Time` or `Duration` since no information about their composition is present within `Typesdict`.

> Note: This does not happen for ROS2 since their message definition always contain `Time` and `Duration`.

# 2) Wrong ROS1 bag split between changing and const data during ROS->Mosaico

 Current split between changing and const data consists in checking whether field name is upper case or not. If it is uppercase it is considered a constant while if it snake_case it is  considered a changing data.
While this is true for ROS2 this is not the case for ROS1. For example the ROS1 `CameraInfo` message type holds data like `D`, `K`,... that are not constant. This leads to an error when transforming these messages to Mosaico.

> Note: This does not happen for ROS2 since `CamerInfo` message definition changed all the upper case fields in snake case

pucciland95:
Fixes #685


mosaico-sdk-py/src/mosaicolabs/protocols/ros2msg.py
```diff
@@ -177,5 +178,14 @@ def convert_ros2msg(msgdef: str, msgtype: str) -> pa.StructType:
                 ])),
             ])
     """
-    types_dict = get_types_from_msg(msgdef, msgtype)
-    return _typesdict_to_schema(types_dict, msgtype)
+    msg_typesdict = get_types_from_msg(msgdef, msgtype)
+
+    # This is necessary since ROS1 msgdef does not contain types like Time and Duration. Therefore,
```
fdicorato: Better formulate this part: "*does not contain types like Time and Duration*". Better say:

'`builtin_interfaces` were introduced in ROS2; within ROS1, `time` and `duration` were built-in types and not part of a separate package.'

mosaico-sdk-py/src/mosaicolabs/ros_bridge/helpers.py
```diff
@@ -91,42 +91,32 @@ def _to_dict(message: Any) -> tuple[Any, Dict[str, Any]]:
     """
```
fdicorato, line 99: I understand the proposed implementation and the rationale behind it. However, it relies on a current rosbags impl, i.e. the generated message modules use `from __future__ import annotations`, so `field.type` is stored as a string (e.g. `"ClassVar[int]"`) rather than an evaluated `typing.ClassVar[...]`.
If a future rosbags version changed how it generates annotations, `field.type` could become an actual typing object and the current string check would no longer work.

A more robust approach we can test is to resolve annotations using `get_type_hints()` (cached per message class) and use `typing.get_origin()` to identify `ClassVar` fields:
```python
# Cache because this is slow, so you evaluate only once per message class-type
@lru_cache(maxsize=None)
def _resolved_type_hints(cls: type) -> Dict[str, Any]:
    return get_type_hints(
        cls,
        globalns=vars(typing), # needed for str-annotations
        include_extras=True,
    )
```
Then:
```python
type_hints = _resolved_type_hints(type(message))

for field_name, field in dataclass_fields.items():
    if field_name == "__msgtype__":
        continue

    field_value = getattr(message, field_name)

    if get_origin(type_hints.get(field_name)) is ClassVar:
        const_dict[field_name] = field_value
    else:
        data_dict[field_name], _ = _to_dict(field_value)
```

This would handle both string annotations and evaluated annotations and avoids relying on the current generated source format. It is To Be Tested, however

fdicorato (changes requested):

pucciland95:
Added a condition to ensure that if rosbags moves from using `annotation` policy this still works:

``` python

if (
      (isinstance(field.type, str) and field.type.startswith("ClassVar["))
      or get_origin(field.type) is ClassVar
  ):
```

fdicorato (approved):

PR #691, opened 2026-07-31
Add arrow_size_bytes parameter 

### LINKED ISSUE #686: Add arrow_size_bytes parameter
`chunk_t` only stores `size_bytes`, the ZSTD-compressed Parquet buffer. 
There is no record of the uncompressed Arrow size of a chunk.

## Change

Add a nullable `chunk_t.arrow_size_bytes BIGINT`, accumulated while writing a chunk with `batch.get_array_memory_size()`, so each chunk carries its uncompressed Arrow footprint.

Use it in `topic_get_stats` to compute `avg_bytes_per_row`, falling back to `size_bytes` per chunk when the value is missing.

LuigiTer:
Added new parameter arrow_size_bytes which accumulate the uncompressed batch size and then calculate the avg_bytes_per_row to compute better optimal batch size.

Close #686


mosaicod/crates/mosaicod-core/src/types/resources.rs
```diff
@@ -278,6 +278,8 @@ impl<M> TopicMetadata<M> {
 pub struct TopicChunksStats {
     pub total_size_bytes: i64,
     pub total_row_count: i64,
+    /// Average row footprint in uncompressed Arrow bytes
+    pub avg_bytes_per_row: f64,
```
DevMattG: Use i64. Being an average we can accept rounding.

mosaicod/crates/mosaicod-db/src/sql/pg_queries/data_catalog.rs
```diff
@@ -214,7 +218,12 @@ pub async fn topic_get_stats(
     let res = sqlx::query!(
         r#"SELECT
             COALESCE(SUM(size_bytes), 0)::BIGINT as "total_size_bytes!",
-            COALESCE(SUM(row_count), 0)::BIGINT as "total_row_count!"
+            COALESCE(SUM(row_count), 0)::BIGINT as "total_row_count!",
+            COALESCE(
+                SUM(COALESCE(arrow_size_bytes, size_bytes))::FLOAT8
+                    / NULLIF(SUM(row_count), 0),
+                0
+            )::FLOAT8 as "avg_bytes_per_row!"
```
DevMattG: even here I'd not return a float, but an integer instead

DevMattG (changes requested):

DevMattG (approved):

PR #697, opened 2026-08-03
docs: fix SDK reference accuracy, fix learn guides, add three new learn guides

fdicorato:
This PR overhauls the Python SDK documentation (`docs/py/docs`, `docs/main/docs/examples`, `docs/main/docs/learn`) and adds three new How-To guides.

- **Correctness fixes**: removed a non-existent `Message.frame_id` field from an example, fixed `.match()` mislabeled as regex instead of glob, added a missing `.outside()` operator to the class-free query table, corrected "Two-way TLS" terminology (the SDK only supports server-authenticated TLS, no mTLS), fixed missing imports and a missing `on_error` policy in the writing guides' Full Examples, and various smaller typos/inconsistencies.
- **Clarity and structure**: reworded overly dense paragraphs, split oversized sections into bullets, reordered content where the "why" appeared after the "how", and removed redundant boilerplate repeated verbatim across code samples.
- **Cross-file de-duplication**: consolidated repeated explanations (wildcard/glob tables, `.Q` proxy description, timestamp semantics, error-policy boxes) into single canonical sources, with the sibling guides linking back instead of restating them.
- **New How-To guides**:
  - *Updating a Sequence* — adding new Topics to an already-ingested Sequence via `SequenceUpdater`, contrasting `client.sequence_update()` (write-only credentials) with `SequenceHandler.update()` (read+write).
  - *Preparing PyTorch Training Batches* — the `DataFrameExtractor` → `SyncTransformer` → windowed `IterableDataset` pipeline for feeding synchronized multi-sensor data into a PyTorch `DataLoader`.
  - *Live Streaming to Foxglove* — streaming a Sequence into a live Foxglove WebSocket session without an intermediate MCAP file, covering well-known Foxglove schemas vs. raw JSON, the `Message`/`Header` ↔ `log_time`/schema-timestamp mapping, replay pacing, server lifecycle, and the `Unmodeled` (class-less) fallback path.

All code snippets were verified against the actual SDK source.

### DIFF
--- a/mosaico-sdk-py/pyproject.toml
+++ b/mosaico-sdk-py/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "mosaicolabs"
-version = "0.6.0-dev"
+version = "0.7.0-dev"
 description = "Python SDK for Mosaico Data-Platform"
 authors = ["Mosaico Labs <info@mosaico.dev>"]
 readme = "README.md"
--- a/docs/main/docs/examples/custom_ontology.mdx
+++ b/docs/main/docs/examples/custom_ontology.mdx
@@ -40,9 +40,9 @@ The custom ontology is loaded as part of the ROS Ingestion example. The full sou
 
 ## Defining the Model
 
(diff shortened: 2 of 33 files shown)


mosaico-sdk-py/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "mosaicolabs"
-version = "0.6.0-dev"
+version = "0.7.0-dev"
```
pucciland95, line 3: Should not this be 0.6.0?

pucciland95 (changes requested):

fdicorato:
replaced by #698

PR #701, opened 2026-08-05
feat(cli): CLI output and various improvements

cavenditti-laif:
Implements most of the CLI improvements outlined in #699:

- Improve how configuration is handled and updated (make profile changes atomic at FS level)
- Add a mosaico doctor command to diagnose and troubleshoot connetion and configuration issues (for humans and agents alike)
- Add guidance for agents: global AGENTS.md + markdown files specific for the CLI, eventually we'll cover the whole repo with appropriate instructions

### DIFF
--- a/docs/main/docs/clients/cli.md
+++ b/docs/main/docs/clients/cli.md
@@ -1,0 +1,58 @@
+---
+title: Mosaico CLI
+sidebar_position: 4
+description: Configure connections, inspect resources, and diagnose Mosaico from the command line.
+---
+
+The `mosaico` command is included with the Python SDK CLI extra:
+
+```bash
+pip install "mosaicolabs[cli]"
+```
+
+## Configure a profile
+
+Interactive setup:
+
+```bash
+mosaico profile add local
+```
+
+Non-interactive setup:
+
+```bash
+mosaico profile add local \
+  --no-interactive \
+  --host localhost \
+  --port 6726 \
+  --default
+```
+
+Profile files are written with owner-only permissions on POSIX systems. Prefer the `MOSAICO_API_KEY` environment variable when credentials should not be stored locally.
+
+## Diagnose a connection
+
+```bash
+mosaico doctor
+```
+
+The command checks the resolved profile, configuration permissions, TLS certificate path, DNS, and TCP connectivity. It never prints the API-key value.
+
+For automation and AI development tools, request structured output:
+
+```bash
+mosaico doctor --output json
+mosaico profile ls --output json
+mosaico sequence ls --output jsonl
+mosaico topic ls --output csv
+```
+
+JSON collection documents include a `schema_version`. Scripts should inspect that field before depending on the document structure.
+
+Skip network checks when validating only local configuration:
+
+```bash
+mosaico doctor --no-network
+```
+
+The command exits non-zero when a required check fails. Warnings, such as overly broad configuration permissions, are reported without hiding otherwise useful diagnostics.
--- a/docs/main/docs/clients/cli.md
+++ b/docs/main/docs/clients/cli.md
@@ -1,6 +1,6 @@
 ---
 title: Mosaico CLI
-sidebar_position: 4
+sidebar_position: 1
 description: Configure connections, inspect resources, and diagnose Mosaico from the command line.
 ---
 
--- a/docs/main/docs/dev/ai_development.md
+++ b/docs/main/docs/dev/ai_development.md
@@ -1,0 +1,36 @@
+---
+title: AI-assisted development
+sidebar_position: 3
+description: Repository guidance and validation expectations for developers working with AI coding agents.
+---
+
+AI coding tools should begin with the repository `AGENTS.md` and then read the closest component-specific instructions. These files define current commands, trust boundaries, and the minimum validation required for a change.
+
+## Recommended workflow
+
+1. Identify the affected Rust crates, Python packages, CLI commands, and documentation.
+2. State the compatibility and security invariants before editing.
+3. Run the smallest relevant tests while iterating.
+4. Update examples and structured-output contracts with the implementation.
+5. Run the change-impact validation described in `AGENTS.md` before handoff.
+
+## Internal specialist agents
+
+Maintainers can use the separate `mosaico-agents` repository for performance, security, consistency, and general maintenance reviews. Those agents share an evidence-first reporting contract:
+
+- distinguish observations from confirmed defects;
+- include commands, workloads, or code paths supporting conclusions;
+- redact credentials and sensitive paths;
+- report validation performed and residual risk;
+- do not claim performance improvement without comparable before/after measurements.
+
+## Machine-readable CLI diagnostics
+
+Use the CLI rather than scraping formatted terminal output:
+
+```bash
+mosaico doctor --output json
+mosaico profile ls --output json
+```
+
+Structured output is intended for automation. Human-oriented tables may change presentation without notice.
--- a/mosaico-sdk-py/src/mosaicolabs_cli/commands/extension.py
+++ b/mosaico-sdk-py/src/mosaicolabs_cli/commands/extension.py
@@ -10,7 +10,12 @@ import typer
 from rich.table import Table
 from typer.core import TyperGroup
 
-from mosaicolabs_cli.utils.config import OutputFormat, console, error_console
+from mosaicolabs_cli.utils.config import (
+    OutputFormat,
+    console,
+    emit_structured_records,
+    error_console,
+)
 
 app = typer.Typer(no_args_is_help=True)
 
@@ -105,6 +110,11 @@ def list_extensions(
     extensions = discover_extensions()
 
     if not extensions:
+        if output in (OutputFormat.JSON, OutputFormat.JSONL):
+            emit_structured_records([], output, "extensions")
+            return
+        if output == OutputFormat.CSV:
+            return
         console.print(
             "[yellow]No external extensions discovered in your $PATH.[/yellow]"
         )
@@ -137,3 +147,14 @@ def list_extensions(
     elif output == OutputFormat.CSV:
         for name, binary_path in extensions.items():
             print(f"{name},mosaico-{name},{binary_path}")
+
+    elif output in (OutputFormat.JSON, OutputFormat.JSONL):
+        records = [
+            {
+                "name": name,
+                "binary": f"mosaico-{name}",
+                "path": binary_path,
+            }
+            for name, binary_path in extensions.items()
+        ]
+        emit_structured_records(records, output, "extensions")
--- a/mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
+++ b/mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
@@ -10,7 +10,7 @@ from mosaicolabs_cli.utils.config import (
     error_console,
     get_config_path,
     load_config,
-    serialize_to_toml,
+    write_config,
 )
 from mosaicolabs_cli.utils.env import DEFAULT_MOSAICO_PORT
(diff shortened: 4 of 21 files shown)


mosaico-sdk-py/src/mosaicolabs_cli/utils/config.py
```diff
@@ -74,6 +79,61 @@ def serialize_to_toml(data: dict) -> str:
...
+
+    return stat.S_IMODE(path.stat().st_mode)
+
+
+def emit_structured_records(
+    records: List[Dict[str, Any]], output: OutputFormat, collection: str
+) -> None:
+    """Emit stable JSON or JSON Lines records without terminal decoration."""
+    if output == OutputFormat.JSON:
+        print(json.dumps({"schema_version": 1, collection: records}, sort_keys=True))
```
DevMattG, line 128: better to avoid magic numbers

DevMattG (changes requested):

docs/main/docs/dev/ai_development.md
```diff
@@ -0,0 +1,36 @@
+---
+title: AI-assisted development
+sidebar_position: 3
```
pucciland95, line 3: The documenation file does not seem to be at position 3 when building the `docs`. However the reason is because in `release_cycle.md` and `testing_and_dev.md` there should be `sidebar_position` rather than `position`. Please, update these in these two cited docs so to have a correct file ordering

docs/main/docs/clients/cli.md
```diff
@@ -0,0 +1,58 @@
+---
+title: Mosaico CLI
+sidebar_position: 4
```
pucciland95: Same as before but now the tag is correct but the numbers are not. I would follow this ordering but I am not sure:

- Mosaico CLI
- Python
- Rust
- C++

what do you think?

pucciland95 (changes requested):

pucciland95 (changes requested):
Leaving some starting comments, however I still need to give a closer look

docs/main/docs/clients/cli.md
cavenditti-laif: No strong opinions on this but it makes sense, I'll change it

mosaico-sdk-py/src/mosaicolabs_cli/utils/config.py
cavenditti-laif: Are you referring to the schema_version?

I wanted a way (mostly for agents) to tell if the returned JSON is aligned with the structure one expects, given the documentation they're referring to. Would you use a different solution or just not include such a field altogether? Maybe relying only on the reported versions of server and CLI?

docs/main/docs/clients/cli.md
```diff
@@ -0,0 +1,58 @@
...
+title: Mosaico CLI
+sidebar_position: 1
+description: Configure connections, inspect resources, and diagnose Mosaico from the command line.
+---
+
+The `mosaico` command is included with the Python SDK CLI extra:
+
+```bash
+pip install "mosaicolabs[cli]"
+```
```
pucciland95, line 11: I would be consistent and mantain Poetry as packet manager. Please consider changing this to:


```suggestion
```bash
poetry install --extras cli
```
```

File: docs/main/docs/clients/cli.md [hunk 5]
```diff
@@ -0,0 +1,58 @@
+---
+title: Mosaico CLI
+sidebar_position: 1
+description: Configure connections, inspect resources, and diagnose Mosaico from the command line.
+---
```
pucciland95, line 5: Consider that now in the main branch there is the new following section about the CLI:

https://github.com/mosaico-labs/mosaico/tree/main/docs/main/docs/cli

that corresponds to this new sections in the main documentation: https://docs.mosaico.dev/cli/commands

In my opinion we should move and integrated this section within the one in main so to avoid repetitions (the CLi should not be within SDKs section). What do you think?

mosaico-sdk-py/src/mosaicolabs_cli/utils/config.py
pucciland95: I think what my collegue was referring to is that assigning `1` to the `schema_version` key has no intuitive meaning for the reader (magic number). 

I get your point on understanding different schema just by checking the `schema_version` attribute however I am not sure it is the best fit. Could you make me an example of a use-case where this is useful?

pucciland95: What about using some kind of hashing as the following:


```suggestion
    schema_version_md5 = hashlib.md5(json.dumps({collection: records}, sort_keys=True).encode("utf-8"))
    schema = json.dumps({"schema_version": schema_version_md5.hexdigest(), collection: records}, sort_keys=True)
```

some pytests should be added to check the behaviour clearly

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
```diff
@@ -219,6 +216,11 @@ def list_profiles(
     config_data = load_config(config_path)
 
     if not config_data:
+        if output in (OutputFormat.JSON, OutputFormat.JSONL):
+            emit_structured_records([], output, "profiles")
+            return
+        if output == OutputFormat.CSV:
+            return
```
pucciland95, line 223: Consider proposed refactoring to better handle this

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
```diff
@@ -259,9 +261,28 @@ def list_profiles(
...
+                    {
+                        "name": p.name,
+                        "host": p.host,
+                        "port": p.port,
+                        "default": p.is_default,
+                        "tls": p.enable_tls,
+                        "api_key_configured": bool(p.api_key),
+                    }
+                )
+        emit_structured_records(records, output, "profiles")
```
pucciland95, line 281: Consider proposed refactoring to better handle this

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
```diff
@@ -259,9 +261,28 @@ def list_profiles(
...
+                records.append(
+                    {
+                        "name": p.name,
+                        "host": p.host,
+                        "port": p.port,
+                        "default": p.is_default,
+                        "tls": p.enable_tls,
+                        "api_key_configured": bool(p.api_key),
+                    }
+                )
```
pucciland95, line 280: What about using:

```suggestion
                records.append(p.to_dict())
```

?

mosaico-sdk-py/src/mosaicolabs_cli/commands/profile.py
```diff
@@ -259,9 +261,28 @@ def list_profiles(
...
+        emit_structured_records(records, output, "profiles")
+
     else:
         error_console.print(
-            f"[bold red]Error:[/bold red] Unsupported output format: '{output}'. Use 'table' or 'csv'."
+            f"[bold red]Error:[/bold red] Unsupported output format: '{output}'."
```
pucciland95, line 285:
```suggestion
            f"[bold red]Error:[/bold red] Unsupported output format: '{output}'. Supported formats are: {[f.value for f in OutputFormat]}"
```

mosaico-sdk-py/src/mosaicolabs_cli/commands/extension.py
```diff
@@ -105,6 +110,11 @@ def list_extensions(
     extensions = discover_extensions()
 
     if not extensions:
+        if output in (OutputFormat.JSON, OutputFormat.JSONL):
+            emit_structured_records([], output, "extensions")
+            return
+        if output == OutputFormat.CSV:
+            return
```
pucciland95, line 117: Consider proposed refactoring to better handle this

mosaico-sdk-py/src/mosaicolabs_cli/commands/extension.py
```diff
@@ -105,6 +110,11 @@ def list_extensions(
     extensions = discover_extensions()
```
pucciland95, line 110: What about creating as for `MosaicoProfile` a `MosaicoExtension` class that gets created by `discover_extensions()` and with a method `to_dict()`. For the moment it could contain things like:

- name
- binary
- path

(Feel free to further modify this)

pucciland95 (changes requested):
I have made some comments to the code.

Futhermore consider the following refactoring to better handle `OutputFormat` (let's discuss about whether you would handle this refactoring differently or something is not clear)

## Refactor: replace format if/else chain with registry-based strategy dispatch

Replacing the growing if/else block that handles `OutputFormat` with a dispatch table. Each format gets its own renderer behind a common interface; renderers self-register and the client looks them up by `OutputFormat` value.

### File layout

**`renderers/base.py`** — the ABC (contract only)
```python
from abc import ABC, abstractmethod
from typing import ClassVar

class OutputRenderer(ABC):
    handled_formats: ClassVar[tuple[OutputFormat, ...]] = ()

    @abstractmethod
    def render(self, data) -> str: ...
```
Note the type is `tuple[OutputFormat, ...]` (variable-length tuple of enum members).

**`renderers/registry.py`** — lookup + registration decorator
```python
_RENDERERS: dict[OutputFormat, OutputRenderer] = {}

def register(cls: type[OutputRenderer]) -> type[OutputRenderer]:
    instance = cls()  # stateless → store a single instance
    for fmt in cls.handled_formats:
        if fmt in _RENDERERS:
            raise ValueError(f"{fmt} already registered")
        _RENDERERS[fmt] = instance
    return cls

def get_renderer(fmt: OutputFormat) -> OutputRenderer:
    try:
        return _RENDERERS[fmt]
    except KeyError:
        raise ValueError(f"No renderer for {fmt}") from None
```

**`renderers/formats.py`** (or one file per format) — concrete renderers
```python
@register
class JsonRenderer(OutputRenderer):
    handled_formats = (OutputFormat.JSON, OutputFormat.JSONL)
    def render(self, data) -> str: ...

@register
class CsvRenderer(OutputRenderer):
    handled_formats = (OutputFormat.CSV,)
    def render(self, data) -> str: ...

# ...TableRenderer, etc.
```

**`renderers/__init__.py`** — force-import the concrete modules so they register
```python
from . import formats  # noqa: F401  (import for registration side effect)
from .registry import get_renderer
```
⚠️ A renderer only registers when its module is imported. If the concrete classes live in files nobody imports, the registry comes up empty — hence the explicit import here.

**Client code** — the whole if/else collapses to:
```python
output = get_renderer(fmt).render(data)
```
`OutputFormat` subclasses `str`, so its members work directly as dict keys — no conversion needed.

Overall folder layout should be:

```
├── output_format.py          # OutputFormat(str, Enum)  ← leaf, imports nothing
└── renderers/
    ├── __init__.py           # imports formats (registration); re-exports get_renderer
    ├── base.py               # OutputRenderer ABC
    ├── registry.py           # _RENDERERS, register, get_renderer
    └── formats.py            # all concrete renderers
```

pucciland95:
A gentle ping @cavenditti-laif

mosaico-sdk-py/src/mosaicolabs_cli/utils/config.py
cavenditti-laif: I wanted an explicit versioning, this way even if updating the schema in a non retrocompatible way, `doctor` would still be able to check it just fine.

The hashing would be less informative, I think going forward with multiple hashes it could become very hard to maintain

cavenditti-laif:
@pucciland95 I'm not totally convinced about the proposal for OutputRenderer, it's cleaner but adds some contraints on the way we use it (think of emitting partial JSONL in the future for whatever reason). A more immediate doubt is about the registration of the OutputRenderer(s), as one may have multiple of them providing handlers for the same formats.

I'll try putting together a solution that addresses these points too.

PR #703, opened 2026-08-06
feat: enhance ROS adapter functionality and error handling

### LINKED ISSUE #702: bug(py): missing ROS msgdef registration in typestore for custom **but adapted** msgtypes
When the an adapter for a non-standard ROS msgtype is available and implemented, the MosaicoLoader skips the registration of the message definition in the typestore. Therefore, when checking 
`if self.typestore.types.get(rosmsg_type) is None:` (mosaico-sdk-py/src/mosaicolabs/ros_bridge/loader.py:1073) the check fails and the topic is always blacklisted.

The correct behavior should be to check for existence of the msgtype in the typestore and, if not present (i.e. custom msgtype) it must be registered by hand via:

```python
# Register new msgdef within typestore
add_types = get_types_from_msg(
   msgdef, msgtype
)
 self.typestore.register(add_types)
```

pucciland95:
Fixes #702

### DIFF
--- a/mosaico-sdk-py/src/mosaicolabs/ros_bridge/adapter_base.py
+++ b/mosaico-sdk-py/src/mosaicolabs/ros_bridge/adapter_base.py
@@ -200,7 +200,7 @@ class ROSAdapterBase(ABC, Generic[T]):
                 - the ros_msg_type or resolved_rosmsg_type are supported but translation is not implemented (NotImplementedError)
         """
         raise NotImplementedError(
-            f"{cls.__name__} does not implement to_ros(). Impossible to encode the Mosaico Message to ROS"
+            f"{cls.__name__} does not implement to_ros(). Unable to encode the Mosaico Message to ROS"
         )
 
     @classmethod
--- a/mosaico-sdk-py/src/mosaicolabs/ros_bridge/adapter_base.py
+++ b/mosaico-sdk-py/src/mosaicolabs/ros_bridge/adapter_base.py
@@ -199,7 +199,9 @@ class ROSAdapterBase(ABC, Generic[T]):
                 - the ros_msg_type or resolved_rosmsg_type are unsupported by typestore (TypeError)
                 - the ros_msg_type or resolved_rosmsg_type are supported but translation is not implemented (NotImplementedError)
         """
-        pass
+        raise NotImplementedError(
+            f"{cls.__name__} does not implement to_ros(). Impossible to encode the Mosaico Message to ROS"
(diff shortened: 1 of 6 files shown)


mosaico-sdk-py/src/mosaicolabs/ros_bridge/adapter_base.py
```diff
@@ -199,7 +199,9 @@ def to_ros(
                 - the ros_msg_type or resolved_rosmsg_type are unsupported by typestore (TypeError)
                 - the ros_msg_type or resolved_rosmsg_type are supported but translation is not implemented (NotImplementedError)
         """
-        pass
+        raise NotImplementedError(
+            f"{cls.__name__} does not implement to_ros(). Impossible to encode the Mosaico Message to ROS"
```
fdicorato: prefer 'Unable' instead of 'Impossible'

mosaico-sdk-py/src/mosaicolabs/ros_bridge/loader.py
```diff
@@ -880,94 +880,162 @@ def __init__(
...
-            if (
-                adapter is not None
-                and adapter.ontology_data_type().__schema_fingerprint__
-                != _compute_schema_fingerprint(t_handler._arrow_schema)
-            ):
-                adapter = None
-
-            resolved_rosmsg_type = (
-                adapter.get_default_ros_msg() if adapter else resolved_rosmsg_type
+        if msgtype is None:
```
fdicorato, line 961: Why raising here? the other adapters factories return None in case of error. Return None here also

mosaico-sdk-py/src/mosaicolabs/ros_bridge/loader.py
```diff
@@ -880,94 +880,162 @@ def __init__(
...
+                or the final fallback does.
+
+        Raises:
+            TypeError: when the topic's ``_ros_`` metadata carries a non-string ``msgtype`` (malformed
+                metadata).
+            RuntimeError: when every hand-written-adapter strategy fails and ``msgtype`` is unknown, so
+                even the unmodeled fallback cannot be created.
+        """
+
+        adapter, resolved_rosmsg_type = (
```
fdicorato: This can result in a runtime error: if all the adapter factories return None, the value cannot be unpacked. Check for Noneness befor unpacking

fdicorato (changes requested):

fdicorato (approved):

PR #715, opened 2026-08-13
refactor(mosaicod)!: target_message_size removed as a user-configurable param

### LINKED ISSUE #696: Remove target_message_size as a user-configurable param
### Problem

Currently we expose two user-configurable params

* `max_grpc_message_size` => Maximum allowed message size in transmission.
* `target_message_size` => Recommended size of a message during transmission.

 that can cause ambiguity and unwanted behaviors if wrongly set.

### Solution

Expose only `max_grpc_message_size` and estimate `target_message_size` internally. This will give us more flexibility in the way we decide to split data batches for transmission (as first implementation consider setting it to half of `max_grpc_message_size`).

`max_grpc_message_size` will be also validated inside a min/max range to avoid wrong values introduce  performance downgrade:
* min: 4 MB (grpc default)
* max: 128 MB

DevMattG:
In this PR `target_message_size` has been removed as a user-configurable param and directly set as half of `grpc_max_message_size` instead.

Min/max value validation has also been introduced for `grpc_max_message_size`

Closes #696

### DIFF
--- a/mosaicod/crates/mosaicod-core/src/params.rs
+++ b/mosaicod/crates/mosaicod-core/src/params.rs
@@ -22,6 +22,9 @@ pub const EPSILON: f64 = 1.0e-06;
 
 pub const MAX_BUFFERED_FUTURES: usize = 8;
 
+pub const GRPC_MSG_MIN_SIZE_BYTES: usize = 4 * 1024 * 1024; // 4MB. Default grpc message size.
+pub const GRPC_MSG_MAX_SIZE_BYTES: usize = 128 * 1024 * 1024; // 128MB
+
 /// Module containing several file extensions
 pub mod ext {
     /// Json file extension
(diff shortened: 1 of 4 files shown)


mosaicod/crates/mosaicod-core/src/params.rs
```diff
@@ -245,10 +276,13 @@ pub fn load_params_from_env(config: ParamsLoadOptions) -> error::PublicResult<()
         .expect("Unable to detect default parallelism, please define MOSAICOD_DEFAULT_PARALLELISM")
         .get();
 
+    let max_grpc_message_size = Param::optional("MOSAICOD_MAX_GRPC_MESSAGE_SIZE", 50 * 1_000_000);
```
LuigiTer: What do you think to define const for all default values?

LuigiTer (approved):

mosaicod/crates/mosaicod-core/src/params.rs
DevMattG: Ok, but I would do it in a separate PR

PR #718, opened 2026-08-13
refactor(mosaicod)!: removed chunks_number and total_bytes from topic_t

### LINKED ISSUE #714: Remove chunks_number and total_bytes from topic_t table
`chunks_number` and `total_bytes` fields used as cache values inside `topic_t` table can be dropped to reduce the cost of maintaining updates to the code in the future without losing so much efficiency.
`chunks_number` can be calculated counting the number of records inside `chunk_t` table for a specific topic, while `total_bytes` can be calculated summing `size_bytes` column.

DevMattG:
I n this PR `chunks_number` and `total_bytes` have been removed from `topic_t` table for they can be easily calculated using info inside `chunk_t` table.

Moreover, topic data_info in `get_flight_info` is now explicitly populated and sent, even if topic is empty.

BREAKING CHANGES:
- now topic's data info is never omitted, even if the topic has 0 chunks.

Closes #714


LuigiTer (approved):
Removing the cached counters from `topic_t` makes sense, but it's worth flagging the runtime cost that comes with it.

`info()` now recomputes the timestamp range through  `compute_data_info` --> `ts_engine.read(...)`,  a directory listing + parquet schema inference + a `min/max` scan over `index_timestamp` on every call, where before it was just columns already loaded with the topic row.

`sequence::info` runs this serially for every topic in the sequence, so the cost scales
with both the number of topics and their chunk count, a sequence made of large topics turns a single `GetFlightInfo` into N scans.

I'd measure `GetFlightInfo` latency on a realistic sequence before merging, so we know what we're trading the cache for.

DevMattG:
Recomputing timestamp ranges inside info() was an error.

I removed `compute_data_info()` completely and moved its logic inside `info()`. Timestamp range is not recompputed every time, but the one already stored inside `topic_t` table is used.

Only chunk stats were to be removed from `topic_t` table.

PR #749, opened 2026-08-31
feat(mosaicod): add ps command to list registered instances 

### LINKED ISSUE #730: Add instance registry to track running server/cleanup processes and cleanup execution status
## Problem

`mosaicod server` and `mosaicod cleanup` are separate processes that can run in any number, on any hosts, against one shared Postgres DB + object store. Multiple `cleanup` processes already coordinate correctness via a Postgres advisory lock (`db::acquire_transaction_lock`, id `777777`) plus the `cleanup_log_t` table, but nothing records *who* is running:

- There's no way to know how many `server`/`cleanup` processes are currently alive, or on which hosts.
- For cleanup specifically, there's no way to tell whether it's idle (waiting for its next scheduled run) or actively executing — and if it's executing, **which specific instance** holds the lock.

This makes it hard to reason about deployments with multiple replicas: e.g. confirming a stuck/crashed cleanup process, or figuring out which host to check logs on when cleanup appears to be hung.

## Proposed solution

Add a lightweight instance registry with heartbeats, attribute each cleanup run to the instance that performed it, and expose both through a new `mosaicod ps` CLI command.

### 1. New `instance_registry_t` table

```sql
CREATE TABLE instance_registry_t
(
    instance_id                     SERIAL PRIMARY KEY,
    kind                            TEXT    NOT NULL,
    hostname                        TEXT    NOT NULL,
    pid                             INTEGER NOT NULL,
    started_unix_tstamp_secs        BIGINT  NOT NULL,
    last_heartbeat_unix_tstamp_secs BIGINT  NOT NULL
);

ALTER TABLE cleanup_log_t
    ADD COLUMN instance_id INTEGER REFERENCES instance_registry_t (instance_id);
```

"Is it alive" is derived purely from heartbeat recency (now - last_heartbeat < staleness_threshold), computed at read time.
Two different `staleness_threshold` are introduced to check the service status:
- one to tell that the service may be down or stopped.
- the other much larger to delete the record from the database.

### 2. Attribute cleanup runs to the instance that performed them

cleanup_log_create gains an instance_id: Option<i32> parameter. "Which instance is running cleanup right now" becomes: read the latest cleanup_log_t row, and if end_unix_tstamp_secs IS NULL, its instance_id is the current executor.

### 3. Registration + heartbeat lifecycle

Both Cleanup::run() (mosaicod-task) and Server::start_and_wait() (mosaicod-grpc) register themselves in the instance registry at startup and run a periodic heartbeat loop (tokio::spawn, ~1min interval) alongside their existing work, tied to the same shutdown signal on failure is treated as fatal (fails startup), consistent with how other DB-dependency failures are handled.

### 4. New CLI command: mosaicod ps

Lists every registered instance (kind / id / host / pid / started / last heartbeat / alive-or-stale) and prints a cleanup status summary:

KIND      ID    HOST            PID      STARTED                LAST HEARTBEAT         STATUS
cleanup   1     host-a          18500    2026-08-21 09:02:15    2026-08-21 10:04:56
server    2     host-a          18422    2026-08-21 09:02:11    2026-08-21 10:05:40    alive
server    3     host-b          991      2026-08-16 09:03:02    2026-08-20 22:10:12

Cleanup: IDLE — last run 2026-08-21 09:10:00 .. 2026-08-21 09:10:03 by instance

## Non-goals

- No CLI flag for the heartbeat interval — hard-coded constants are enough.
- No gRPC health/admin endpoint — this is a DB-backed CLI-only feature for now.
- Assumes reasonably NTP-synced clocks across hosts when comparing heartbeat recency from a different reader host.

DevMattG:
## Summary                                                                                                                                                                                                                       
  - Add a `mosaicod ps` subcommand that lists registered `mosaicod` instances (servers and cleanup routines) and summarizes the cleanup routine's current status, backed by a new `instance_registry` table and periodic heartbeats.                                                                                                                                                                                            
  - Derive each instance's liveness (`alive` / `stale` / `dead`) from time since its last heartbeat, and show UPTIME per instance.                                                                                                 
  - Add `-a`/`--all` to include `dead` instances in the listing (hidden by default).                                                                                                                                               
  - Add `-v`/`--verbose` to show the STARTED column and switch LAST HEARTBEAT to an exact UTC timestamp; by default LAST HEARTBEAT shows a relative time ("3m 20s ago", "14h 12m ago"), or "-" for dead instances.

Closes #730


mosaicod/crates/mosaicod-core/src/types/instance.rs
```diff
@@ -0,0 +1,50 @@
...
+
+/// Identifies the kind of long-running `mosaicod` process registered in the instance registry
+/// (see `mosaicod ps`).
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+pub enum InstanceKind {
+    /// A `mosaicod server` process.
+    Server,
+    /// A `mosaicod cleanup` process.
+    Cleanup,
+}
```
cabba, line 25: `StoreOptimizer` is missing

mosaicod/crates/mosaicod-task/src/cleanup.rs
```diff
@@ -11,6 +11,65 @@ const TO_DELETE_MARKER_FILE_NAME: &str = "TO_DELETE";
 const DEFAULT_TIME_INTERVAL: u32 = 86400;
 const DEFAULT_RETENTION_DURATION: u32 = 86400;
 
+/// Reads the local hostname, falling back to `"unknown"` if it can't be determined.
+fn local_hostname() -> String {
```
cabba: This function is replicated also in `mosaicod/crates/mosaicod-grpc/src/server.rs`, consider move the function in a shared crate.

mosaicod/crates/mosaicod-task/src/cleanup.rs
```diff
@@ -11,6 +11,65 @@ const TO_DELETE_MARKER_FILE_NAME: &str = "TO_DELETE";
...
+}
+
+/// Periodically refreshes `instance_id`'s heartbeat in the instance registry, and opportunistically
+/// purges long-expired entries, until `shutdown` is cancelled.
+///
+/// This is intentionally its own small loop (rather than folded into [`Cleanup::run`]'s loop):
+/// the cleanup routine's own schedule is driven by `time_interval` (which can be as sparse as
+/// once a day), while the heartbeat needs a much tighter, fixed cadence to be a useful liveness
+/// signal.
+async fn instance_heartbeat_loop(db: db::Database, instance_id: i32, shutdown: CancellationToken) {
```
cabba: This function is replicated also in mosaicod/crates/mosaicod-grpc/src/server.rs, consider move the function in a shared crate.

cabba (changes requested):

mosaicod/crates/mosaicod-commands/src/print.rs
```diff
@@ -117,3 +114,191 @@ pub fn store_display_name(store: &store::StoreRef) -> String {
...
+
+    println!("{}", header.join(" "));
+
+    let instances: Vec<&db::InstanceRegistryRecord> = instances
+        .iter()
+        .filter(|i| show_dead || !matches!(i.status(), db::InstanceStatus::Dead))
+        .collect();
+
+    if instances.is_empty() {
+        println!("{}", "No registered instances found.".dimmed());
```
LuigiTer: Is it correct in the case we have dead instances but `--all` is not set? 
I would distinguish the case empty registry and hidden case (e.g `x hidden istances, use --all to show them`)

LuigiTer (changes requested):

mosaicod/crates/mosaicod-commands/src/print.rs
DevMattG: What do you prefer between:
- removing completely the println (as `docker ps` does)
- changing it into something like "No instances. Try with --all to see more." (and just "No instances" when `--all` is enabled)

LuigiTer: I suggest the second option, in this way the user knows there is the option `--all`

DevMattG:
With the last commit I had to add a check inside store optimizer to stop its loop even when already working on a topic, due to the long wait that otherwise could be experienced after sending a SIGINT. Interrupting the optimization of a topic should not be a problem: the newly created data folder is left orphan and thus the cleanup routine will take care of it, while any change to the DB is dropped if not yet committed.

mosaicod/crates/mosaicod-commands/src/print.rs
```diff
@@ -117,3 +114,196 @@ pub fn store_display_name(store: &store::StoreRef) -> String {
...
+    };
+
+    let by_instance = log
+        .instance_id
+        .and_then(|id| instances.iter().find(|i| i.instance_id == id))
+        .map(|i| format!(" by instance {} ({})", i.instance_id, i.hostname))
+        .unwrap_or_default();
+
+    match log.end_datetime() {
+        None => {
```
LuigiTer, line 291: ps will display always `running` if cleanup for some reason will be killed or panic. 
Probably you can check it easily with `istance_id` if the instance is registered or in a dead state

LuigiTer (changes requested):

mosaicod/crates/mosaicod-commands/src/print.rs
DevMattG: I added a `INTERRUPTED` state to indicate when a cleanup exits unexpectedly in the middle of a run. It is triggered only if the instance is `DEAD` and not `STALE`

LuigiTer (approved):

mosaicod/crates/mosaicod-core/src/params.rs
```diff
@@ -43,6 +43,18 @@ pub const DEFAULT_STORE_SECRET_KEY: &str = "";
 pub const DEFAULT_STORE_ACCESS_KEY: &str = "";
 pub const DEFAULT_STORE_OPTIMIZER_MEMORY_POOL_SIZE: usize = 0;
 
+// Instance registry (see `mosaicod ps`). Not configurable: these are cheap, low-stakes
+// background-loop knobs not requiring a CLI flag or env var.
+
+/// Interval, in seconds, between instance-registry heartbeats emitted by long-running
+/// `mosaicod` processes (server, cleanup).
+pub const INSTANCE_HEARTBEAT_INTERVAL_SECS: u32 = 30;
```
cabba, line 51: Question: we plan to expose these options in the config file?

cabba (approved):

PR #763, opened 2026-09-02
feat(py): Add schema metadata consistency and broken URLs tests

### LINKED ISSUE #762: feat(py): Add new tests: check broken links in documentation and check consistency in received schema metadata (sdk->mosaicod->sdk)

fdicorato:
Closes #762

### DIFF
--- a/mosaico-sdk-py/src/testing/unit/test_links.py
+++ b/mosaico-sdk-py/src/testing/unit/test_links.py
@@ -1,0 +1,146 @@
+"""Checks every README.md in the repo and every Markdown/MDX file under docs/ for broken URLs.
+
+Only http(s) links are checked (relative/anchor links are intentionally out of scope).
+A link is considered broken when it definitively responds with 404/410.
+"""
+
+from __future__ import annotations
+
+import re
+import socket
+from concurrent.futures import ThreadPoolExecutor
+from dataclasses import dataclass
+from pathlib import Path
+from urllib.error import HTTPError, URLError
+from urllib.parse import urlsplit
+from urllib.request import Request, urlopen
+
+import pytest
+
+REPO_ROOT = Path(__file__).resolve().parents[4]
+DOCS_DIR = REPO_ROOT / "docs"
+
+EXCLUDED_DIR_NAMES = {
+    ".git",
+    ".venv",
+    "venv",
+    "node_modules",
+    "__pycache__",
+    ".docusaurus",
+    ".pytest_cache",
+    ".mypy_cache",
+    ".ruff_cache",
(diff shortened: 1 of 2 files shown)


mosaico-sdk-py/src/testing/unit/test_links.py
```diff
@@ -0,0 +1,146 @@
...
+
+def iter_markdown_files() -> list[Path]:
+    files: set[Path] = set()
+
+    for path in REPO_ROOT.rglob("README.md"):
+        if not _is_excluded(path):
+            files.add(path)
+
+    if DOCS_DIR.is_dir():
+        for pattern in ("*.md", "*.mdx"):
```
pucciland95, line 71:
```suggestion
        ACCEPTED_FILES = ("*.md", "*.mdx")

        for pattern in ACCEPTED_FILES
```

mosaico-sdk-py/src/testing/unit/test_links.py
```diff
@@ -0,0 +1,146 @@
...
+            problems.append(BrokenLink(link.source, link.line, link.target, error))
+    return problems
+
+
+@pytest.mark.skipif(not _has_internet(), reason="no network access")
+def test_external_links_are_reachable():
+    problems = find_external_link_problems()
+    assert not problems, "Broken external links found:\n" + "\n".join(
+        str(p) for p in problems
+    )
```
pucciland95, line 146: Consider this only if you consider making this a test that **does not** breaks local and Git Hub CI but just warns the user

```suggestion
    if problems:
        pytest.xfail(
            "Broken external links found:\n" + "\n".join(str(p) for p in problems)
        )
```

To actually see the output `pytest` needs to be launched with `-ra` option (and so add to all `poetry run pytest` `-ra` in `scripts/tests`)

pucciland95 (changes requested):
Python SDK refers to main documentation through `https` links and [main online documentation](https://docs.mosaico.dev/) is updated only when the release happens. Because of this, during each development cycle from one version to the next, whenever Python SDK refers to not (yet) deployed documentation `test_external_links_are_reachable` will always fail breaking the CI and therefore blocking the PRs. 

I propose to make `test_external_links_are_reachable` a non-breaking test (`xfail`) as explained in the review comments.

What do you think?

mosaico-sdk-py/src/testing/unit/test_links.py
fdicorato: @DevMattG, considering the comments of @pucciland95, I think that your proposal to move this test into docs dir, can be the right solution

PR #789, opened 2026-09-10
feat!: new 'info' do_action endpoint replacing 'version'

DevMattG:
This PR introduces the new `info` do_action endpoint which replaces and extends the old `version`.

Besides sharing the server version, it is used to expose also some of the server's configuration params to the clients
(e.g. `max_grpc_message_size` and `target_message_size`). 

Closes #731

PR #805, opened 2026-09-16
ci: ci now covers all integration tests

### LINKED ISSUE #804: Extend full stack tests
To solve the issue detailed in dicussion [801](https://github.com/mosaico-labs/mosaico/discussions/801) we will extend the workflow file `.github/workflows/ci.yml` with `--all` option.

cabba:
Closes #804


pucciland95:
@cabba could you please review the PyTest changes?

pucciland95:
Commenting to highlight why PyTests `test_connect_enables_gzip_options`, `test_connect_enables_gzip_with_level_options` and `test_connect_disables_gzip_by_default` are breaking.

These tests complain that when the client makes a `info` action the response returned is empty, raising a `ConnectionError` and making the test fail. Bug was introduced in #789, where `wait_for_connection()` has been changed. 

Before #789  `wait_for_connection()` makes a `version` action, disregarding the response: 

``` python
client.do_action(fl.Action("version", body))
```

After #789  `wait_for_connection()` changed that line with:

``` python
action_results = client.do_action(fl.Action(action_name, body))
```

checking that `action_results` contains information about the server. 
Since the three cited integration tests use a `mock` of the client (there is no actual connection with the server), all the action call return `None`. 
In the old code, the return value was not checked and therefore the tests were passing. Conversely the new code checks that the `action_result` is not `None` breaking the tests since the connection is mocked.

To fix this issue, since the client is `patched`, it makes sense to `patch()` also the `wait_for_connection()` function

pucciland95 (approved):

PR #807, opened 2026-09-17
refactor(mosaicod): unify TimestampRange across marshal and ext crates

### LINKED ISSUE #806: Consolidate duplicated TimestampRange/timestamp-window types across crates
## Problem

Currently there are at least four separate representations of "a `[start, end)` timestamp window" scattered across the codebase, each with its own conversion boilerplate:

- `mosaicod_marshal::flight::TopicAppMetadataTimestamp` (`start_ns: i64, end_ns: i64`)
- `mosaicod_marshal::flight::FilterTimestampRange` (`start_ns: u64, end_ns: u64`, with its own `validate()` and `TryFrom<types::TimestampRange>`/`From<&Self>` impls)
- `mosaicod_marshal::actions::responses::ClusterTimestampRange` (`start_ns: u64, end_ns: u64`)
- `mosaicod_ext::arrow_filter::Cluster` — raw `start_ns: u64` / `end_ns: u64` fields instead of a range type

This duplication means:
- Validation logic (`start >= end` → bad request) is implemented independently in `FilterTimestampRange::validate` and is *missing* for the `topic filter_clusterize` action, so a request with `start_ns == end_ns` currently succeeds instead of being rejected like `get_flight_info` already does for sequences.
- `mosaicod_core::types::TimestampRange::is_empty()` only checks `start > end`, not `start >= end`, which is inconsistent with the half-open `[start, end)` semantics documented on the struct and with how `FilterTimestampRange` treats an empty window.
- `arrow_filter::extract_timestamps` reinterprets the Arrow `Int64Array` backing slice as `&[u64]` via an `unsafe` pointer cast, justified by a comment asserting timestamps are "always non-negative" — this is an unnecessary unsafe block working around the fact that `Cluster` (and the clustering algorithm) uses `u64` instead of the natural `i64` timestamp type used everywhere else.

## Proposed changes

- [x] Add a single shared `TimestampRange` type in `mosaicod-marshal` (serde-friendly, wrapping/convertible to `mosaicod_core::types::TimestampRange`) and use it everywhere a wire-format timestamp window is needed, replacing `TopicAppMetadataTimestamp`, `FilterTimestampRange`, and `ClusterTimestampRange`.
- [x] Fix `mosaicod_core::types::TimestampRange::is_empty()` to use `start >= end` per the documented half-open interval semantics, and add `PartialEq`/`Eq` derives so it can be compared directly in tests.
- [x] Validate `timestamp_range` in `topic filter_clusterize` (`crates/mosaicod-grpc-flight/src/endpoint/actions/topic.rs`) the same way `get_flight_info` does, rejecting empty ranges (including `start == end`) with `core::Error::bad_timestamp_range` before doing any ontology/clustering work.
- [x] Change `mosaicod_ext::arrow_filter::Cluster` to hold a `types::TimestampRange` instead of separate `start_ns`/`end_ns: u64` fields, and update the clustering/intersection logic (`cluster_stream`, `intersect_cluster_streams`) to operate on `i64` timestamps natively.
- [x] Remove the `unsafe` pointer-cast in `extract_timestamps` — return `&[i64]` directly from the Arrow array instead of reinterpreting it as `&[u64]`.
- [x] Update `intersect_cluster_streams`'s saturating-arithmetic to work over signed `i64` (e.g. `checked_add_unsigned`/`saturating_sub_unsigned`) now that clusters carry signed timestamps.

## Why

Reduces duplicated conversion/validation code across crates, closes a validation gap where `filter_clusterize` silently accepted empty timestamp windows, and removes an `unsafe` block that exists only because clustering used the wrong integer signedness.

DevMattG:
- Replace TopicAppMetadataTimestamp, FilterTimestampRange, and ClusterTimestampRange with a single mosaicod_marshal::TimestampRange used across flight metadata, action requests/responses, and clustering code paths.
- Change TimestampRange::is_empty() to use start >= end, matching the documented half-open [start, end) semantics, and derive PartialEq/Eq for it.
- Reject empty timestamp ranges (start >= end) in topic filter_clusterize the same way get_flight_info already does.
- Switch mosaicod_ext::arrow_filter::Cluster to hold a types::TimestampRange instead of raw start_ns/end_ns: u64 fields, and drop the unsafe pointer cast in extract_timestamps by operating on i64 timestamps natively.

Closes #806


mosaicod/crates/mosaicod-ext/src/arrow_filter.rs
```diff
@@ -125,23 +125,22 @@ where
...
-                        start_ns: t,
-                        end_ns: t,
+                        timestamp_range: types::TimestampRange::between(t.into(), t.into()),
                         id,
                     });
                 }
                 (Some(curr), Some(prev)) => {
-                    if (t - prev) <= clustering_dt_ns {
-                        curr.end_ns = t;
+                    debug_assert!(t >= prev);
```
cabba, line 133: 👍

cabba (approved):

mosaicod/crates/mosaicod-core/src/types/time.rs
```diff
@@ -134,9 +134,9 @@ impl TimestampRange {
         self.start.is_unbounded() && self.end.is_unbounded()
     }
 
-    /// Check if the timestamp range is empty (i.e. start > end)
+    /// Check if the timestamp range is empty (i.e. start >= end)
     pub fn is_empty(&self) -> bool {
-        self.start > self.end
+        self.start >= self.end
```
LuigiTer, line 139: This update prevents creating instantaneous events where the start and end times match. Is this the behavior that we want?

LuigiTer (approved):

mosaicod/crates/mosaicod-core/src/types/time.rs
cabba: We consolidated the time interval definition as $[t_{start}, t_{end})$.
So the interval $[x, x)$ makes no sense and is considered empty.

PR #816, opened 2026-09-22
bug(py): update as_py() calls to avoid dict to tuple conversion during streaming

### LINKED ISSUE #813: `pa.dict` are streamed as list of tuples when streaming from Server
When streaming a `Serializable` object containing within the `PyArrow` schema a field of type `PyArrow.map_` it is interpreted as a `list` of `Tuple` instead of a `dict`. For example, loading:

``` python
{
   'start_point': {'latitude': 55.07, 'longitude': 27.686, 'altitude': 139.0}, 
   'moving_target': {'latitude': 15.07, 'longitude': 27.6861, 'altitude': 39.0}, 
   'home': {'latitude': 40.0, 'longitude': 8.0, 'altitude': 100.0}
}
```

when streamed from the server returns:

``` python
[ 
  ('start_point', {'latitude': 55.07, 'longitude': 27.686, 'altitude': 139.0}) 
  ('moving_target', {'latitude': 15.07, 'longitude': 27.6861, 'altitude': 39.0}),     
  ('home', {'latitude': 40.0, 'longitude': 8.0, 'altitude': 100.0})
] 
```

Despite having this `PyArrow` schema:

``` python
   map<string, struct<latitude: double not null, 
                      longitude: double not null, 
                      altitude: double not null>>
```

 This creates a type missmatch between what it is expected (a `dict`) and what it is actually returned (`list` of `Tuple`).

pucciland95:
As explained in #813,  `dict` get cast to `list` of `Tuple` during data streaming. 

To solve this, when using `as_py()` to convert  `PyArrow` data coming from the server into a Python `dict`, it is necessary to specify the parameter `maps_as_pydicts="lossy"`, so nested `dict` remain Python `dict` and are not turned into a `list` of `tuple`. See [here](https://arrow.apache.org/docs/python/generated/pyarrow.StructScalar.html#pyarrow.StructScalar.as_py) for more reference.

An integration test that covers this behavior has been added.

Fixes: #813


fdicorato:
From the linked source:

 "If ‘lossy’, whenever duplicate keys are detected, a warning will be printed. The last seen value of a duplicate key will be in the Python dictionary. If ‘strict’, this instead results in an exception being raised when detected."

Why not setting 'strict' for checking data consistency?

pucciland95:
Known issues:

- Impossible to query `Any` data value since in PyArrow they are saved as `pa.bytes()`
- protobuf dictionary keys with `int` type cannot fall outside `[-2^53, 2^53 - 1]` interval since they may be converted to string from protobuf `MessageToDict`

